PR #10, opened 2021-08-05
Read number cell

Schiznitz:
Hello,
I found that the worksheet reader just simply skip number value cell which type_value is "n" and this fix works for me, hope it helps.
Regards,


MathNya:
Hi Schiznitz.
Thank you so much. :)
We will reflect the points you pointed out in the next version update.

PR #13, opened 2021-08-31
Fixed column index handling for columns >= Z

8191:
Parsing columns Z, AA, ..., ZZZ did not work as expected.
Calculating the string representation of column indices > 25 did not work as expected.

I've also written some unit tests to avoid this in future. ;-)


MathNya:
thank you for all your help.
I've committed to it.

PR #14, opened 2021-10-19
writer: support io::Write

matt-duch:
Hi! Would you be interested in this change?
This modifies the writer to allow passing in an arbitrary `io::Write`, and writes directly into the final zip archive, rather than into an intermediate temp directory.


MathNya:
Hi matt-duch.
Thank you so much!
I had a chance to look at the revision you proposed.
They were very smart and excellent.
I will merge them.

PR #15, opened 2021-10-21
writer: fix path generation

matt-duch:
This is a bugfix. I didn't realize the old approach actually didn't prepend the `dir` argument into the path, resulting in broken paths (duplicate prefixes prepended). I relied on the integration tests, which apparently didn't catch this issue. I've tested files generated with this code using OpenOffice.


MathNya:
Thank you for your frequent support.
I will merge this one as well.

PR #17, opened 2021-12-09
reader/xlsx: support io::Reader

matt-duch:
So I've made a similar change to my previous PR for making the writer accept any io::Write trait, but this time for the reader (allowing any io::Read). There's a number of formatting changes in other files from an automated refactor where cargo fmt ran automatically. I hope that's ok. if not, let me know, and I may try to to the refactor again, but it would take at least an hour or two.
This also decouples the code entirely from using the filesystem, both for the read and write API, as well as removing the dependence on the temp directory on a file system.

Happy to give you more details on the changes involved.


MathNya:
Thank you for your continued help.
I have checked the changes.
There are no problems, and we will merge them.

PR #21, opened 2022-01-31
[*.rs] rustfmt

SamuelMarks:
Also found these warnings: [rust.log](https://github.com/MathNya/umya-spreadsheet/files/7973580/rust.log)

Some are trivial, like you should remove the `license-file` from your `Cargo.toml`


MathNya:
Hi, SamuelMarks.

Thanks for the suggestion.
I hadn't used rustfmt before.
I'll commit to it.

PR #30, opened 2022-03-11
use hashmap to get shared string table id

### LINKED ISSUE #28: writing string type value is still pretty slow
for example
https://github.com/MathNya/umya-spreadsheet/blob/39f9ee5fedb2bdac7a644d683763dd270d6556d2/src/structs/shared_string_table.rs#L56
I think it's better to use hashmap.

cstkingkey:
Fix #28


MathNya:
We will merge them.
Thank you for your response.

PR #33, opened 2022-03-12
use Hashmap

cstkingkey:
cargo test --release time drops from 101s to 8s.

just FYR. more work need to be done.


MathNya:
We have confirmed that there are no problems, so we will merge them.

PR #34, opened 2022-03-13
Branchname

MathNya:
Fixed for "use Hashmap #33".


MathNya:
We have confirmed that there are no problems, so we will merge them.

PR #37, opened 2022-04-04
add  method to remove cell

cstkingkey:


MathNya:
Thank you for everything.
We will merge as soon as we are sure there is no problem.

PR #39, opened 2022-04-05
Fixes XLOOKUP function in excel

ubamrein:
When using a `XLOOKUP` function, which results in multiple values, excel seems to put some xml attributes on the formula element (`t=arr` and `ref=START:END`). This pull request just parses all attributes and puts all of them back as they were present when the spread sheet was opened. Adding those attributes let's Excel correctly show the XLOOKUP. There is still some difference to the original Excel version though, which probably needs further investigation.

Further, this PR fixes a problem, when a `empty_tag` was not false. In this case, a `c` start-tag would have been emitted, without the corresponding end tag (since the end tag was within the if clause).

### DIFF
--- a/src/structs/cell.rs
+++ b/src/structs/cell.rs
@@ -236,6 +236,13 @@ impl Cell {
         &self.cell_value.get_formula()
     }
 
+    pub(crate) fn get_formula_attributes(&self) -> Vec<(&str, &str)> {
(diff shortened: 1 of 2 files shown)


MathNya:
Thank you for the pull request.
We will merge them as soon as the content is confirmed.

PR #48, opened 2022-04-07
use typed cell value and parse cell value lazily

cstkingkey:
for reference


KyGost:
> not sure about the relationships between value, rich text and formula.
> 
> `pub fn get_value(&self) -> &str { match &self.value { Some(v) => { return v; } None => {} } match &self.rich_text { Some(v) => { return v.get_text(); } None => {} } "" }`
> 
> it seems to be mutually exclusive.

What do you mean by this?

How come you are framing it around `get_value -> &str`?
Would these changes not imply `get_value -> Value`?

cstkingkey:
> > not sure about the relationships between value, rich text and formula.
> > `pub fn get_value(&self) -> &str { match &self.value { Some(v) => { return v; } None => {} } match &self.rich_text { Some(v) => { return v.get_text(); } None => {} } "" }`
> > it seems to be mutually exclusive.
> 
> What do you mean by this?
> 
> How come you are framing it around `get_value -> &str`? Would these changes not imply `get_value -> Value`?

it's the original code

PR #53, opened 2022-04-08
Use Clippy

KyGost:
I appreciate that this is a massive change and as such I don't expect it to be merged but I figured I'd show it.

These changed were largely done with [`cargo clippy`](https://github.com/rust-lang/rust-clippy).
> A collection of lints to catch common mistakes and improve your [Rust](https://github.com/rust-lang/rust) code.

I like using clippy as it is useful for improvements, cleanups and consistency.


MathNya:
Thank you for your suggestion.
We have merged them.

Issue #52, opened 2022-04-08
Panic


KyGost:
I'll try to get more details if I can, for now all I can give is that when opening a sheet I'm getting:
```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', .../umya-spreadsheet-0.5.3/src/structs/stylesheet.rs:160:18
```

KyGost:
nb. not fixed by `0.6.0`:
```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', .../umya-spreadsheet-0.6.0/src/structs/stylesheet.rs:160:18
```

KyGost:
Offending line:
https://github.com/MathNya/umya-spreadsheet/blob/3a385edfcf16769878e4ca4130dbcc35dcc924d9/src/structs/stylesheet.rs#L160

KyGost:
I've fixed this for now by changing it from:
```rust
.unwrap()
.clone();
```
To:
```rust
.cloned()
.unwrap_or_default();
```

KyGost:
Oh...
But now I'm getting
```
thread 'main' panicked at 'This Worksheet is Not Serialized. Please exec to read_sheet(&mut self, index: usize);', /home/kyran/Documents/GitHub/umya-spreadsheet/src/structs/spreadsheet.rs:447:21
```


I'm going to try to change the method from lazy to normal.

KyGost:
After changing it from lazy I get this:
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ParseFloatError { kind: Empty }', /home/kyran/Documents/GitHub/umya-spreadsheet/src/structs/double_value.rs:26:52
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

KyGost:
Lots of layers....
```
thread 'main' panicked at 'index out of bounds: the len is 0 but the index is 0', /home/kyran/Documents/GitHub/umya-spreadsheet/src/reader/xlsx/vml_drawing.rs:35:29
```

MathNya:
Thank you for the information.
We will check the contents.

PR #54, opened 2022-04-08
Handle Panics

KyGost:
Resolves: #52 

Simple fixes, likely incorrect.


MathNya:
This event was not reproduced.
The fix was fine and has been merged.

Issue #41, opened 2022-04-06
Remove Onig and use Regex instead


ubamrein:
When building for a target, for which no c/c++ compiler toolchain is present, `umya-spreadsheet` cannot be built. Especially, when trying to target wasm, it gets annoying to handle the compiler. This is caused by the regular expression engine used (`onig`), which is a wrapper around a c library.

Further, the zip dependency needs to disable `bzip2` support (or at least allow it via e.g. feature flags), and instead relay on flate2, which is a pure rust implementation of some deflate based compression algorithms.

I have a version locally tested to work with `regex` instead of `onig` and was also able to use it in the web-browser via wasm-pack. If this is something you are interested in, I can make a PR.

MathNya:
Thank you for contacting us.
We are very interested in your proposal.
If you don't mind, could you please send us a PR?

PR #63, opened 2022-04-11
remove onig and use regex instead

ubamrein:
PR for https://github.com/MathNya/umya-spreadsheet/issues/41

This PR removes `onig` and relays on `regex` instead. Further only the `flate2` dependency with `rust_backend` feature is used (`bzip2` disabled), such that it can be built for wasm.


MathNya:
Thank you very much!
We will merge them.

PR #65, opened 2022-04-13
Use fancy-regex and reduce number of regex compilations

ubamrein:
Sorry the PR yesterday was not uptodate with your upstream, so it seems there were two more adjustments. I fixed it now. Sorry for the inconvenience.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@ description = "umya-spreadsheet is a library written in pure Rust and read and w
 [dependencies]
 quick-xml = { version = "0.17", features = [ "serialize" ] }
 zip = {version = "0.5.6", default-features = false, features = ["deflate"]}
-regex = "1.3.7"
+# regex = "1.3.7"
 md-5 = "0.10.1"
 lazy_static = "1.4.0"
 thousands = "0.2.0"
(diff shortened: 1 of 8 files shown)


cstkingkey:
cannot pass tests as unsupported Regex pattern

cstkingkey:
and it's better to convert more strings into cow< str> than vice versa.

ubamrein:
> cannot pass tests as unsupported Regex pattern

yes I saw that in my IDE, it still builds though and since the other PR was merged I thought it is just a random lint rule which is activated in my place.

> and it's better to convert more strings into cow< str> than vice versa.

True that! So I relabel this PR as WIP and will adjust the api to minimize cloning.

ubamrein:
So I realised that there are quite some regex which use look aheads. I changed from `regex` to `fancy_regex` which is built on top of `regex` and as such also pure rust (hence building to wasm e.g. is very simple). 

Regarding the Cows, I had trouble getting it to work with the lifetimes, since all the regex functions reborrow from the input. I removed the calls to `to_string`, but introduced variable shadowing, so that in a future refactoring, one could actually figure the life times out :D 

Further, I used lazy_static to avoid compiling the same Regex multiple times. Where a regex was not needed, I used `std::string::replace` to avoid compiling and using regular expressions.

This time I also ran the tests, and everything seems to work. 

Please tell me if you want anything changed :)

MathNya:
Thank you very much for your great support.
I have no additional requests. Please allow me to merge them if you like.

PR #88, opened 2023-01-17
Set column and row index to one

tomgroenwoldt:
When adding a comment to a xlsx sheet the process panics due to the index currently starting at 0 by default. This doesn't work as indices start at 1 in the sheet, so it makes sense to adjust the default value.


MathNya:
hi. [tomgroenwoldt](https://github.com/tomgroenwoldt)
Thank you for the PR.
Excellent suggestion.
We will merge them.

tomgroenwoldt:
Thanks for the quick response. This is an awesome crate! :)

PR #92, opened 2023-01-31
Fix XML space preserve attribute

kjh618:
Currently the XML space preserve attribute is inserted only if there's a whitespace at the start of the string. Change this to insert the attribute when there's a whitespace at the end too.


MathNya:
Thank you for the PR.
We have no problem with the revisions and will merge them.

PR #93, opened 2023-02-04
[Update] deprecate `*_by_column_and_row*` methods

john-dc252:
## Changes:

  - deprecate `*_by_column_and_row*` methods
  - deprecate `index_from_coordinate_simple`
  - add struct `CellCoordinates` for cell coordinates
  - make methods that take address strings as parameters generic and accept `Into<CellCoordinates>` instead

> The changes listed above will allow having just one method/function
> for things that need an address string or column and row index.
> This means methods like `get_cell` do not need a separate `get_cell_by_column_and_row` anymore.
> e.g.

```rust
// both these calls are now valid
worksheet.get_cell("C2");
worksheet.get_cell((3, 2));
```


MathNya:
Wow, thanks for the cool suggestion.
I will merge them.

Issue #74, opened 2022-05-17
improve reading speed


cstkingkey:
run cargo flamegraph --test integration_test -- read_large_string
the result show that (1) umya_spreadsheet::helper::coordinate::index_from_coordinate and (2) umya_spreadsheet::structs::cells::Cells::set_fast are the main time consumers.

(1) is related to regex and format.
(2) hashbrown::raw::RawTable<T,A>::reserve_rehash is the main time consumer.   set the capacity of the HashMap properly can improve the speed. If there is  a way to read the total cells in the worksheet in advance, it will be the best.

MathNya:
Thank you for your report.
I will check the areas you mentioned.

PR #94, opened 2023-02-06
[Update] refactor column address conversion code

john-dc252:
might help with #74


MathNya:
Thank you for giving me a hand.
We will merge this information as well.

PR #96, opened 2023-02-08
[Update] refactor index_from_coordinate

john-dc252:
## Changes:

  - change return type of `index_from_coordinate` to tuple
  `(u32, u32, bool, bool)` (the lock flags were previously converted into `&str` first)


MathNya:
We will merge the PRs we receive.
We will release it as soon as possible after we have confirmed that it works.

PR #100, opened 2023-02-11
[Update] make `CellValue` internal

john-dc252:
## Changes:

  - change visibility of `CellValue` to `pub(crate)`
  - remove internally unused methods that return `CellValue`

### DIFF
--- a/src/structs/cell.rs
+++ b/src/structs/cell.rs
@@ -26,17 +26,12 @@ pub struct Cell {
     hyperlink: Option<Hyperlink>,
 }
 impl Cell {
-    pub fn get_cell_value(&self) -> &CellValue {
+    pub(crate) fn get_cell_value(&self) -> &CellValue {
         &self.cell_value
     }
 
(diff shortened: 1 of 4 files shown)


john-dc252:
Since mutators and accessors are already available through `Cell`, would it not be better to exclude the `CellValue` type from the public API?

MathNya:
[john-dc252](https://github.com/john-dc252)
> Since mutators and accessors are already available through Cell, would it not be better to exclude the CellValue type from the public API?

Sorry for the late reply.
The CellValue API, I think it would be useful to have.
Like when preparing a CellValue and applying it to multiple Cells.
If this is the case, I feel that Cell's set_value_from_string etc. is unnecessary.
As for the system, I want to keep it as simple as possible.
Hmmm, I will consider this a little.

c-git:
If I could add my 2 cents. I'm not sure how much value would be added by removing it. I do agree it is possible to trip up new users but I think a manual or something like that explaining how to do common things would better manage helping new users and provide content for the AI's to use to better help users.

PR #109, opened 2023-03-01
editing cells in a new sheet is not working as the corresponding row …

cstkingkey:
…is not cretaed.


MathNya:
We will merge this one as well.

Issue #110, opened 2023-03-07
Unable to modify and save cell value in umya_spreadsheet 0.9 version


BYC30:
Issue: Unable to modify and save cell value in umya_spreadsheet 0.9 version

## Description:

The provided code is unable to modify and save cell value in umya_spreadsheet 0.9 version. The code works fine in version 0.8.7 where it can modify A1 cell value and save it in the file "aaa2.xlsx". However, when the code is run in version 0.9, it fails to modify A1 cell value, and hence cannot save it in the file "aaa2.xlsx".

```rust
extern crate umya_spreadsheet;

fn main(){

    let path = std::path::Path::new("./aaa.xlsx");
    let mut book = umya_spreadsheet::reader::xlsx::read(path).unwrap();
    
    let sheet = book.get_sheet_by_name_mut("Sheet1").unwrap();
    let cell = sheet.get_cell_mut("A1");
    // work on 0.87
    // cell.set_value("test".to_string());
    // work on 0.9
    cell.set_value_from_string("test");
    let path = std::path::Path::new("./aaa2.xlsx");
    let r = umya_spreadsheet::writer::xlsx::write(&book, path);
    println!("{:?}", r);
}
```

## Expected Behavior:

The code should be able to modify the A1 cell value and save it in the file "aaa2.xlsx" when run on umya_spreadsheet 0.9 version, similar to how it works in version 0.8.7.

## Steps to Reproduce:

Install umya_spreadsheet 0.9 version.
create empty excel aaa.xlsx
Run the provided code on version 0.9.

## Actual Behavior:

The code runs without any error, but the A1 cell value is not modified and hence not saved in the file "aaa2.xlsx".

vonkruel:
The fix for this is in, but as of now there's no release that includes the fix.  You can build the library locally from master until there's a new release.  That's what I'm doing, but it's not a solution that's free of its own problems by any means.

When API changes have been made, it's a particularly bad time for the library to be left in an unusable state like this.  At least in my own application, this bug causes my code to generate nothing but empty sheets.  If I hadn't changed my code to use the new API, it would have been a simple matter to change the dependency back to the previous version.  I hope you can put an end to this situation by making a new release soon.

BYC30:
I understand the situation.
It's unfortunate that the fix hasn't been released yet.
I have been using the previous version of the library and have found it to be very reliable.

MathNya:
We apologize for any inconvenience caused.
We will release 0.9.1 in the near future.

PR #113, opened 2023-03-10
fix: fix styles not getting applied

john-dc252:
this fixes an issue similar to #110


MathNya:
Thanks for the PR.
We will merge it.

PR #122, opened 2023-06-09
Change write_write method signatures

ggodlewski:
I was trying to achieve getting spreadsheet into memory using following code:

```
let mut data: Vec<u8> = Vec::new();
umya_spreadsheet::writer::xlsx::write_writer(&spreadsheet, &mut data)?;
Ok(data)
```

It was failing so:
* I removed unnecessary `Seek` trait
* and moved `mut` into parameter

After that it works perfectly


MathNya:
@ggodlewski 
Thanks for the PR.
We have confirmed that there are no problems with the revisions and have merged them.

PR #133, opened 2023-09-26
Skip setting numbering format if format is not found

popen2:


MathNya:
@popen2 
Thanks for the PR.
We will merge them.

PR #134, opened 2023-09-27
Fix joining paths when target is absolute

popen2:
This happens when reading an Excel file with comments, their paths start with `/xl/comments` which crashes in `src/structs/raw/raw_file.rs:67`.


MathNya:
@popen2 
Thank you for this as well.
We will merge them.

PR #135, opened 2023-10-21
support for dataValidation.operator

patrickomatic:
I noticed that `operator` support wasn't implemented (and I need it).  I went off of these docs but it was really hard to find anything concrete so please let me know if I'm missing anything: https://learn.microsoft.com/en-my/dotnet/api/documentformat.openxml.spreadsheet.datavalidationoperatorvalues?view=openxml-2.8.1


MathNya:
@patrickomatic 
Thanks for the RP.
We have no problem with the modifications and will merge them in.

PR #136, opened 2023-10-30
Fix drawing error

patrickomatic:
Hi, it's me again - I was running into this error on a spreadsheet that was written by Excel:

```
thread 'main' panicked at /Users/patrick/umya-spreadsheet/src/structs/drawing/spreadsheet/worksheet_drawing.rs:350:61:
index out of bounds: the len is 0 but the index is 0
```

I'm not sure why it doesn't have (or wasn't able to parse) the expected OLE Objects data structures, but when I debug there it was an empty vector.  So I just made the code handle it better but I'm assuming that the drawing gets stripped out of the final sheet.

I also fixed some clippy warnings in the first 3 commits.


MathNya:
@patrickomatic 
Thank you very much.
We will merge the PRs.

PR #137, opened 2023-11-01
Tables

vonkruel:
Hi,

CCD-Analytics has implemented minimal support for data tables in USS (with both read & write support).  We didn't implement the full spec, but we hope you'll agree it's a reasonable start and does provide utility as-is.  I thought about explaining the API, but I'm sure browsing the changes & new files will tell you everything.  If you find this isn't suitable for merging, please let me know what needs to be done.  We'll definitely prefer not to maintain our own fork of USS.

Thanks for your time & your work on this very useful project.

Best,
Adam


MathNya:
@vonkruel 
We have been waiting for you.
No problem with the updated content.
We will merge them.
Thank you for the great PR.

PR #138, opened 2023-11-01
Fix swapped attributes

patrickomatic:
I was working on something else and noticed the code here has these two swapped. Reading the docs it seems like `xfId` should map to `format_id` and `builtinId` to `builtin_id`.  I also found a second place where the vertical/horizontal attrs are flipped


MathNya:
@patrickomatic 
I had not noticed this.
Thanks for the PR.

PR #140, opened 2023-11-06
Fix fmt errors

patrickomatic:
This fixes how master is currently broken, the code just needed `cargo fmt --all` run on it (most people just hook it up to their editor so it runs automatically).  

I also added a badge to the README showing the current build status and some small instructions for those who want to contribute


MathNya:
@patrickomatic 
Thank you for all the details.
We will merge them.

PR #142, opened 2023-11-09
a simple bugfix for the data table reader

vonkruel:
I was browsing the diffs & noticed a little boo-boo which is fixed here.

Thanks for accepting our tables PR.  It's great to have that in.  👍

Best,
Adam


MathNya:
@vonkruel 
Thanks for the PR.
We will merge them.

Issue #143, opened 2023-11-13
xml escape character "&" is corrupting sheetname on write


agentjill:
in `book.get_sheet_by_name_mut("Sheet2").unwrap().get_cell_mut("A1").set_value("TEST1");` 

if the sheet name contains xml escape characters for example say `Alice&Bob` then it is not detecting the `Worksheet`.

On inspection the  ''Alice&Bob'' is being processed as `Alice&amp;Bob`, that would mean that xml escape character `&amp;` is not properly read as `&`.

calling sheet by `Alice&amp;Bob `is saving the workbook as `Alice&&amp;amp;Bob`

MathNya:
@agentjill
Thank you for your report.
We will address this in the next version update.

agentjill:
This was addressed in #145.

PR #145, opened 2023-11-15
Fix for sheet name containing xml escape sequence

agentjill:
As reported in issue #143 , the following is my approach towards solving this. Kindly look into this

### DIFF
--- a/src/reader/xlsx/workbook.rs
+++ b/src/reader/xlsx/workbook.rs
@@ -2,6 +2,7 @@ use crate::xml_read_loop;
 
 use super::driver::*;
 use super::XlsxError;
+use quick_xml::escape;
 use quick_xml::events::Event;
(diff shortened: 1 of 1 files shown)


agentjill:
Another approach is to use `worksheet.set_name(convert_character_reference(&name_value));` instead of explicitly calling quick-xml again.

MathNya:
@agentjill 
Thank you for your consideration.
I think this modification is smart and good.
Please release Draft and we will merge it.

https://github.com/MathNya/umya-spreadsheet/pull/145/files#diff-f0f4f74f345f6f1fa8641d024663826eefc687ba2f3bea48d3415e455f366adfR44

agentjill:
@MathNya thanks for the consideration.

MathNya:
@agentjill 
We will merge them.

PR #149, opened 2023-11-27
added additional attribute "hidden"

agentjill:
In some sheets the some cells are set to "**hidden**" as part of protection to make their internal value (mostly formula) not accessible. The support for the same is added. Kindly look for change in the function `pub(crate) fn get_hash_code(&self) -> String {...}` for required change which I feel is necessary and not sure as to how to go about it

### DIFF
--- a/src/structs/protection.rs
+++ b/src/structs/protection.rs
@@ -12,7 +12,6 @@ use writer::driver::*;
 pub struct Protection {
     locked: BooleanValue,
     hidden: BooleanValue,
-
 }
 
 impl Protection {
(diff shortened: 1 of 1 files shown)


MathNya:
@agentjill 
Thanks for the PR.
Please implement get_hash_code() like this.
```rust
    pub(crate) fn get_hash_code(&self) -> String {
        format!(
            "{:x}",
            md5::Md5::digest(format!("{}{}", &self.locked.get_hash_string(), &self.hidden.get_hash_string()))
        )
    }
```

agentjill:
@MathNya as requested the `get_hash_code(&self)` is also added and kindly approve PR if all is OK

MathNya:
@agentjill 
We will merge them.

PR #160, opened 2023-12-09
basic fixes

agentjill:


MathNya:
@agentjill 
Thank you as always.
We will integrate them.

PR #161, opened 2023-12-12
moved `const` strings into separate variables for readability.

agentjill:
@MathNya when reading the source, I felt there are excess number of magic strings floating around. Therefore, I moved them into corresponding variables in naming convention as of "openpyxl". If you approve this, I will be planning to move them into a separate file.


MathNya:
@agentjill 
Thank you very much.
We will merge them.

agentjill:
@MathNya kindly suggest necessary changes

MathNya:
@agentjill 
Thank you for your suggestion.
const looks better if we put them in one file.
It looks like the following.

[src\helper\const.rs].
pub const VML_NS: &str = "http://schemas.openxmlformats.org/officeDocument/2006/relationships/vmlDrawing";

By doing so, it could be used for writer\xlsx\content_types.rs etc.

Thank you for your confirmation.

agentjill:
@MathNya I have done refactoring as much as possible. I agree there is still pending refactor in some few places. I feel this point is good enough for your review to merge. Kindly look and suggest necessary changes.

agentjill:
@MathNya, the work is mostly done apart one or two instances of refactoring. Kindly merge this PR so that additional work can be done on my part regarding milestones of this repo.

MathNya:
@agentjill 
I have verified that the modifications are correct and will merge them in.

Thank you very much for your help this year.
Have a happy new year!

PR #166, opened 2024-01-03
minor changes to logic

agentjill:
The following changes are done:

- Formatted `cargo.toml` file alphabetically and with consistent format.
- Replaced sensible occurrences of panic!() with assert!() for better readability.
-In `src/reader/xlsx.rs` the return is already `XlsxError` therefore it doesn't make sense to `unwrap()` there. Therefore, the error due to read is passed to caller function.
- in `src/structs/drawing/charts/plot_area.rs` some modification to minimize redundant lines.
- in `src/structs/numbering_format.rs` it made sense to use inbuilt method `expect()` rather than elaborate checking done.
- in `src/structs/range.rs` as the `panic!()` was already preventing the case where `coordinate_collection` is other than 1 or 2. It didn't made sense to check again. Hence removed it. Also as it is already known that `coordinate_collection` is always has length of atleast one. Hence removed redundunt checks.
- In `src/writer/xlsx/shared_strings.rs` small changes to logic to prevent redundant allocation.
- In `src/structs/worksheet.rs` small logical changes to done and removed elaborate checks for simpler logic.
- In `src/structs/spreadsheet.rs` the function return type of several functions changed from `Result` type to `Option` type as the only logical reason why getting a sheet with a certain criterion may fail is because it is not found. Therefore reducing the allocation of static string in the binary.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,27 +11,27 @@ description = "umya-spreadsheet is a library written in pure Rust to read and wr
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-quick-xml = { version = "0.31.0", features = [ "serialize" ] }
-zip = {version = "0.6.6", default-features = false, features = ["deflate"]}
-md-5 = "0.10.6"
-lazy_static = "1.4.0"
-thousands = "0.2.0"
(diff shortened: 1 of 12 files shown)


MathNya:
@agentjill 
Thanks again.
This is great PR. I have verified that the updates are correct and will merge them in.

PR #169, opened 2024-01-08
Some changes

agentjill:
@MathNya, in theme of making changes as felt necessary, In this PR I have made the same.

- General chages
- In `src/helper/coordinate.rs`, the major change was for logical checking in function `adjustment_insert_coordinate(..)`, additional check not necessary as guaranteed by compiler because of function signature being u8. More details mentioned at points individually.
- In `src/structs/coordinate.rs` a `to_string()` method was added to struct `Coordinate` because the name `get_coordinate()` didnt match to its function. The use of later was changed to wherever mentioned in the crate but not removed due to breaking change.
- In `src/writer/driver.rs` the inbuilt function `extend_attributes(..)` was used in place of the manual implementation.
- In `src/helper/html.rs` there are some changes in pub trait potentially leading to breaking change. I felt this necessary because functions returning bools weren't named so to reflect same.
- changed the `match` statements where only one path was followed to `if let` syntax for clarity purposes. Also, inverted several of the `if` statements for clarity purposes.

### DIFF
--- a/src/helper/coordinate.rs
+++ b/src/helper/coordinate.rs
@@ -103,27 +103,27 @@ pub fn coordinate_from_index_with_lock(
 ) -> String {
     format!(
         "{}{}{}{}",
-        if is_lock_col == &true { "$" } else { "" },
+        if *is_lock_col { "$" } else { "" },
         string_from_column_index(col),
-        if is_lock_row == &true { "$" } else { "" },
+        if *is_lock_row { "$" } else { "" },
         row
     )
 }
 
 pub(crate) fn adjustment_insert_coordinate(num: &u32, root_num: &u32, offset_num: &u32) -> u32 {
-    let mut result = *num;
-    if (num >= root_num && offset_num > &0) || (num < root_num && offset_num < &0) {
-        result += offset_num;
+    if (num >= root_num && offset_num != &0) {
+        num + offset_num
+    } else {
+        *num
     }
-    result
 }
 
 pub(crate) fn adjustment_remove_coordinate(num: &u32, root_num: &u32, offset_num: &u32) -> u32 {
-    let mut result = *num;
-    if (num >= root_num && offset_num > &0) || (num < root_num && offset_num < &0) {
-        result -= offset_num;
+    if (num >= root_num && offset_num != &0) {
+        num - offset_num
+    } else {
+        *num
     }
-    result
 }
 
 pub type CellIndex = (Option<u32>, Option<u32>, Option<bool>, Option<bool>);
@@ -161,8 +161,7 @@ impl From<String> for CellCoordinates {
 
 impl From<&str> for CellCoordinates {
     fn from(value: &str) -> Self {
-        let coordinate_upper = value.to_uppercase();
-        let (col, row, ..) = index_from_coordinate(coordinate_upper);
+        let (col, row, ..) = index_from_coordinate(value.to_uppercase());
         CellCoordinates::new(col.unwrap(), row.unwrap())
     }
 }
--- a/src/helper/html.rs
+++ b/src/helper/html.rs
@@ -51,12 +51,12 @@ pub fn html_to_richtext_custom(
 fn read_node(node_list: &Vec<Node>, parent_element: &Vec<HfdElement>) -> Vec<HtmlFlatData> {
     let mut result: Vec<HtmlFlatData> = Vec::new();
 
-    if node_list.len() == 0 {
+    if node_list.is_empty() {
         return result;
     }
 
     let mut data = HtmlFlatData::default();
-    data.element.append(&mut parent_element.clone());
+    data.element.extend_from_slice(parent_element);
 
     for node in node_list {
         match node {
@@ -76,18 +76,18 @@ fn read_node(node_list: &Vec<Node>, parent_element: &Vec<HfdElement>) -> Vec<Htm
 
                 let mut elm: HfdElement = HfdElement::default();
                 elm.name = element.name.to_string();
-                let mut attributes: HashMap<String, String> = HashMap::new();
-                for (name, value) in &element.attributes {
-                    attributes.insert(
-                        name.clone(),
-                        if value.is_some() {
-                            value.as_ref().unwrap().to_string()
-                        } else {
-                            String::from("")
-                        },
-                    );
-                }
-                elm.attributes = attributes;
+
+                elm.attributes = element
+                    .attributes
+                    .iter()
+                    .map(|(name, value)| {
+                        (
+                            name.clone(),
+                            value.as_ref().map(|v| v.to_string()).unwrap_or_default(),
+                        )
+                    })
+                    .collect();
+
                 elm.classes = element.classes.clone();
                 data.element.push(elm);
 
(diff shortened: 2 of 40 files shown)


src/helper/coordinate.rs
```diff
@@ -103,27 +103,27 @@ pub fn coordinate_from_index_with_lock(
...
         string_from_column_index(col),
-        if is_lock_row == &true { "$" } else { "" },
+        if *is_lock_row { "$" } else { "" },
         row
     )
 }
 
 pub(crate) fn adjustment_insert_coordinate(num: &u32, root_num: &u32, offset_num: &u32) -> u32 {
-    let mut result = *num;
-    if (num >= root_num && offset_num > &0) || (num < root_num && offset_num < &0) {
```
agentjill, line 115: the `offset_num` is never less than zero because it is an `u8` therefore this always be false. hence the other part of or is not needed.

src/helper/coordinate.rs
```diff
@@ -103,27 +103,27 @@ pub fn coordinate_from_index_with_lock(
...
+        num + offset_num
+    } else {
+        *num
     }
-    result
 }
 
 pub(crate) fn adjustment_remove_coordinate(num: &u32, root_num: &u32, offset_num: &u32) -> u32 {
-    let mut result = *num;
-    if (num >= root_num && offset_num > &0) || (num < root_num && offset_num < &0) {
```
agentjill, line 123: As mentioned above, similar logic for `offset_num` is provided to remove additonal or branch

src/helper/html.rs
```diff
@@ -51,12 +51,12 @@ pub fn html_to_richtext_custom(
...
     let mut result: Vec<HtmlFlatData> = Vec::new();
 
-    if node_list.len() == 0 {
+    if node_list.is_empty() {
         return result;
     }
 
     let mut data = HtmlFlatData::default();
-    data.element.append(&mut parent_element.clone());
+    data.element.extend_from_slice(parent_element);
```
agentjill, line 59: using inbuilt methods for readability

src/helper/html.rs
```diff
@@ -113,57 +113,51 @@ fn make_rich_text(html_flat_data_list: &Vec<HtmlFlatData>, method: &AnalysisMeth
...
-        let mut underline: bool = method.underline(html_flat_data);
-        let mut superscript: bool = method.superscript(html_flat_data);
-        let mut subscript: bool = method.subscript(html_flat_data);
-        let mut strikethrough: bool = method.strikethrough(html_flat_data);
+        let mut is_bold: bool = method.is_bold(html_flat_data);
+        let mut is_italic: bool = method.is_italic(html_flat_data);
+        let mut is_underline: bool = method.is_underline(html_flat_data);
+        let mut is_superscript: bool = method.is_superscript(html_flat_data);
+        let mut is_subscript: bool = method.is_subscript(html_flat_data);
+        let mut is_strikethrough: bool = method.is_strikethrough(html_flat_data);
```
agentjill, line 121: named as such to better reflect the function

src/helper/html.rs
```diff
@@ -186,176 +180,98 @@ pub struct HfdElement {
     classes: Vec<String>,
 }
 impl HfdElement {
-    pub fn get_by_name(&self, name: &str) -> bool {
-        let mut result = false;
-        if self.name == name {
-            result = true;
-        }
-        result
+    pub fn has_name(&self, name: &str) -> bool {
```
agentjill, line 183: As this function only returns a bool therefore name appropriately and is a **breaking** change.

src/helper/html.rs
```diff
@@ -186,176 +180,98 @@ pub struct HfdElement {
...
 
-    pub fn get_by_class(&self, class: &str) -> bool {
-        let mut result = false;
-        for cls in &self.classes {
-            if class == cls {
-                result = true;
-            }
-        }
-        result
+    pub fn contains_class(&self, class: &str) -> bool {
```
agentjill, line 193: named as such because the function earlier name does not convey what it does. This is a **breaking** change.

src/helper/html.rs
```diff
@@ -186,176 +180,98 @@ pub struct HfdElement {
...
-    fn underline(&self, html_flat_data: &HtmlFlatData) -> bool;
-    fn superscript(&self, html_flat_data: &HtmlFlatData) -> bool;
-    fn subscript(&self, html_flat_data: &HtmlFlatData) -> bool;
-    fn strikethrough(&self, html_flat_data: &HtmlFlatData) -> bool;
+    fn is_bold(&self, html_flat_data: &HtmlFlatData) -> bool;
+    fn is_italic(&self, html_flat_data: &HtmlFlatData) -> bool;
+    fn is_underline(&self, html_flat_data: &HtmlFlatData) -> bool;
+    fn is_superscript(&self, html_flat_data: &HtmlFlatData) -> bool;
+    fn is_subscript(&self, html_flat_data: &HtmlFlatData) -> bool;
+    fn is_strikethrough(&self, html_flat_data: &HtmlFlatData) -> bool;
```
agentjill, line 208: The trait function names changed to describe the functionality and potentially a **breaking** change.

src/structs/cell.rs
```diff
@@ -201,23 +201,20 @@ impl Cell {
...
-        for value in value_list {
+
+        value.split('\n').fold(0f64, |mut acc, value| {
             let mut point = 0f64;
             for chr in value.chars() {
-                let mut clen = chr.len_utf8() as f64;
-                if clen > 1f64 {
-                    clen = 1.5;
-                }
+                let clen = if chr.len_utf8() > 1 { 1.5 } else { 1.0 };
```
agentjill, line 209: here if `clen` if checked for greater than 1 then as `len_utf8()` can give only values 1to4 inclusive then this logic is similar to previous.

src/structs/coordinate.rs
```diff
@@ -7,6 +7,18 @@ pub struct Coordinate {
     column: ColumnReference,
     row: RowReference,
 }
+
+impl ToString for Coordinate {
```
agentjill, line 11: added `to_string()` to `Coordinate`

src/writer/xlsx.rs
```diff
@@ -105,92 +105,86 @@ fn make_buffer(spreadsheet: &Spreadsheet, is_light: bool) -> Result<std::vec::Ve
...
+            worksheet
+                .get_raw_data_of_worksheet()
+                .write(&worksheet_no, &mut writer_manager)?;
         }
         worksheet_no += 1;
     }
 
     // Objects associated with worksheets
-    let mut worksheet_no: i32 = 1;
+    let mut worksheet_no = 0;
```
agentjill, line 128: slight change for code readability. No change in function workings

src/helper/coordinate.rs
agentjill, line 115: the `offset_num` is never less than zero because it is an `u8` therefore this always be false. hence the other part of or is not needed.

agentjill (commented):
kindly look into my reasoning.

MathNya:
@agentjill 
Thank you for the great correction.
We have confirmed that there are no problems and will merge them in.

PR #170, opened 2024-01-10
Changing Reference of Option to Optional Reference

agentjill:
**Alert this is a breaking change**

- The `&Option<T>` in the library has been changed to `Option<&T>` similarly for mutable case for better memory management and general uniformity in the code base.
- Changed the unnecessary `match` statements to `if let` syntax for better readability.
- Removed unnecessary `let _ =` and making code more idiomatic.
- Also changed unnecessary `if let Some(_)` pattern to `is_some()` function
- Also sprinkled in some general improvements.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,15 +12,15 @@ description = "umya-spreadsheet is a library written in pure Rust to read and wr
 
 [dependencies]
 aes = "0.8.3"
-ahash = "0.8.6"
-base64 = "0.21.5"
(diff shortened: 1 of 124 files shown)


agentjill:
@MathNya if you want to merge this PR kindly squash the commits into a single commit, for better git history.

MathNya:
@agentjill 
Thanks again for the tremendous corrections.
We will merge them.

PR #174, opened 2024-02-06
some improvements/optimizations

agentjill:
some changes on following

- Unnecessary cloning of values removed and replaced with alternative strategy wherever felt necessary.
- Some unnecessary string invocation removed and used string slices to save on some space.
- Added additional definitions to `const_str.rs`.
- changed code to idiomatically use options and results.
- nested code simplified by adding guarded if 
- in `cells.rs` and `rows.rs` the logic changed to remove cloning by using `take` function as it is just changing the key value of a hashmap with itself.
- simplified `update_pivot_cache` function in `src/structs/spreadsheet.rs`


MathNya:
@agentjill 
Thank you for everything.
We will merge them.
After adding some fixes, we will release it as version 1.2.0 around the end of this week.

PR #175, opened 2024-02-12
Update CI

patrickomatic:
* Updates `actions/checkout@v3` which gets rid of a warning message about Node versions 
Screenshot 2024-02-12 at 6 52 38 PM
* Combines the 3 separate build steps into one so they share the same build (quicker)


MathNya:
@patrickomatic 
Thanks for the PR.
We will merge them.

PR #179, opened 2024-02-14
Allow to change `codeName` attribute in `workbookPr`

nikvoid:
Add a way to change workbook `codeName` attribute when saving spreadsheet. This allows to use dumped `vbaProject` code in `Workbook` object as-is.

Example with default `codeName` :
image
Notice that there are `Workbook________` (imported code) and `Workbook________1` (new) objects. And imported code is in first one, which is not active, so user needs to manually copy code. 

If we change codeName to `Workbook________`:
image
No  `Workbook________1` entry and code now is working out of the box.


MathNya:
@nikvoid 
Thanks for the PR.
We will merge them.

PR #180, opened 2024-02-14
some improvement changes

agentjill:
@MathNya
In tradition with the previous PR's this PR is also a slight refactor and slight improvement.

- Some `&Option<T>` changed to `Option<&T>` similarly for mutable case. I think these were spared during the previous purge. (As always a **disclaimer for breaking change**)
- In `src/structs/string_value.rs`, `get_value_string()` was renamed to `get_value_str()` to properly indicate its return type.(**NOT** a breaking change)
- Some idiomatic changes
- In `src/structs/spreadsheet.rs` the function `get_pivot_caches()` changed so that it is shortcircuiting. Noticable speed bump is observed.

### DIFF
--- a/src/structs/cell_style.rs
+++ b/src/structs/cell_style.rs
@@ -56,7 +56,7 @@ impl CellStyle {
     pub(crate) fn write_to(&self, writer: &mut Writer<Cursor<Vec<u8>>>) {
         // cellStyle
         let mut attributes: Vec<(&str, &str)> = Vec::new();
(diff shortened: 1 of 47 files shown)


agentjill:
@MathNya I will be taking a break with commits to this repo. Is there a way other than gitter where I can contact you. An email or other instant messager perhaps?

MathNya:
@agentjill 
Thank you for your continued support.
I will merge them.

My email address is now available. It is listed here.
https://github.com/MathNya

PR #186, opened 2024-05-02
Add `js` feature flag

fabianboesiger:
Adds the `js` feature flag and passes it down to the `getrandom` dependency.


MathNya:
@fabianboesiger 
Thank you.
This is an ideal fix.
We will merge it.

PR #191, opened 2024-05-31
fix(color): off-by-one for `INDEXED_COLORS` access

RoloEdits:
Current implimentation sets the value right from the `color` tags value, e.g `<color indexed="2"`, which is already zero-indexed.

https://github.com/MathNya/umya-spreadsheet/blob/3c34cb8d09c242c214aa33640378fd096f641e6e/src/structs/color.rs#L212-L215

The `- 1` would then access the previous indexes color if it was `<color indexed="1.."`, or return `None` if it was `<color indexed="0"`.

The flipside to this is change is that the implimentation for `set_value` no longer needs to add 1. The tests for this have been modified for this change.

I added a single test with red color "RED" in `A1`
image

which results in an `styles.xml` with:
```xml
<font>
    <sz val="11.000000" />
    <color indexed="2" />
    <name val="Calibri" />
    <scheme val="minor" />
</font>
```


MathNya:
@RoloEdits 
Thank you very much.
We will merge three PRs.

PR #196, opened 2024-06-15
Bumping dependencies to newer versions

agentjill:
hi @MathNya , just a PR bumping dependencies to newer versions, also added are some changes to avoid unnecessary string creation.


agentjill:
the tests failed are due to commit "8c93b18dbb103d969ea6b5692cd7c5fe7469e4d6". Kindly look into the same

MathNya:
@agentjill 
It has been a while.
Thanks for the PR.
I am aware that the error is due to my update.
I will merge them.

PR #197, opened 2024-06-19
Minor changes

agentjill:
The `StringVal` struct methods are named properly to reflect their functionality and `get_value()` is changed to meet expected scenarios. Also include some minor changes


MathNya:
@agentjill 
Thank you for your continued support.
We will merge them.

PR #198, opened 2024-06-22
`StringValue` utilization and other minor changes

agentjill:
Changed all struct fields of `Option<String>` to `StringValue` which is internal making the codebase more streamlined, and other minor changes. **Note: Maybe a breaking change due to `pub fn` surface changes for API's to be more flexible**


MathNya:
@agentjill 
Thank you very much.
We will merge them.

PR #202, opened 2024-06-27
Changing `CellRawValue` Variant to be more meaningful

agentjill:
@MathNya, The Variant of Enum `CellRawValue` is given as `Null`. I find it to be confusing to the user. Therefore, I changed the name of the default item of the `CellRawValue` to be `Empty` to be more meaningful. 

In `src/structs/cell_value.rs` & `src/structs/cell.rs` the function `set_value()` if given text "NULL" would store nothing into a cell, which can also be achieved when passed "". Therefore, I feel this should be rationalised. 

**Note: This would be a breaking change**

### DIFF
--- a/src/helper/crypt.rs
+++ b/src/helper/crypt.rs
@@ -93,7 +93,7 @@ pub fn encrypt_revisions_protection(password: &str, workbook_protection: &mut Wo
     workbook_protection.remove_revisions_password_raw();
 }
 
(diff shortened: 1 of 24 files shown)


agentjill:
@MathNya, I need to clarify some things, what is the difference between `Str` Variant and `String` variant in `CellRawValue` Enum. Kindly also explain the need for `Inline` Variant when inline strings are stored as strings anyway.

PR #211, opened 2024-07-18
fix: char len cause panic

attila-lin:


MathNya:
Thanks for the PR.
I will merge it.

PR #212, opened 2024-07-22
fix the bug with number format

hackers267:
fix the bug with number format


MathNya:
@hackers267 
Thanks for the PR.
we will merge it.

PR #214, opened 2024-07-22
fix: Error Message

agentjill:


MathNya:
@agentjill 
Thanks for the PR.
There seems to be an error, but from looking at the source, it looks fine, so I will merge it.

Issue #199, opened 2024-06-24
Proposal: Feature flags for Additional functionalities


agentjill:
In line with #158 & #165, features such as image,drawing,chart support can be moved to a single/multiple feature flags as per the broad consensus of the community. This would reduce the size, execution times and memory footprints of the binary and can in general be good for flexibility of the system. **Note: Same is being implemented with [`Calamine`](https://github.com/tafia/calamine) library which can be investigated for replication.**

MathNya:
@agentjill 
Thank you for your suggestion.
I agree with your response to features.
Let's definitely do it.

xamgore:
Absolutely worth it. As written in [a guide](https://matklad.github.io/2021/09/04/fast-rust-builds.html#Read-The-Lockfile):

> This is undoubtedly a very useful feature to have for some applications, but in the context of rust-analyzer we don’t need it. Simple env_logger-style filtering would be enough.
>
> Once you identify a similar redundant dependency, it’s usually enough to tweak features field somewhere, or to send a PR upstream to make non-essential bits configurable.

If we check the compilation time with `cargo +nightly build -p YOUR_PACKAGE --timings --release`, we may see that these image encoders/decoders take a huge time, even though in my project I don't really them. The image to be inserted into a cell is just a `include_bytes!("some.jpg")` and is already encoded.

<img width="250" alt="image" src="https://github.com/user-attachments/assets/8ae875c1-677f-4b8b-bdf4-c407bd4c61fc"> <img height="200" alt="image" src="https://github.com/user-attachments/assets/5862c892-93f4-4945-a120-4d167abd508f">

MathNya:
@xamgore 
Thanks for the useful information!
I will try to proceed with this correspondence as soon as possible.

xamgore:
The issue can be fixed in many ways.

The first is to proxy all the `image` codec features through `umya-spreadsheet`. So if someone needs `png` they will turn on the feature `umya-spreadsheet = { version = "*",  features = ["image/png"] }`. Most of the time, people insert only one concrete, static image, so... that's enough for them.

On the other hand, when people insert an image from a previously defined set (for example, organization stamps which are meant to be printed on paper), usually those images have the same dimensions, to prevent bloating. Which means a new method could be made:

```rust
#[cfg(feature = "image")]
pub fn new_image(&mut self, path: &str, marker: MarkerType) {
    let path = std::path::Path::new(path);
    let img = image::open(path).unwrap();
[...]

PR #223, opened 2024-09-11
Speed up image stuff

xamgore:
Partially contributes to #199, as it makes three main things:

- replaces reading from a `File` with a reading from a `BufReader` to increase speed
- replaces `image::open()` with `image::image_dimensions()` to increase speed
- makes `image` crate optional to increase compilation speed

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "umya-spreadsheet"
-version = "2.0.2"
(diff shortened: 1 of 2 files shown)


MathNya:
@xamgore 
Thanks for the PR.
I am getting some kind of error, but I will merge it.
(I will fix the error if I can resolve it on my end.)

PR #226, opened 2024-09-17
[WIP] Fix for formula being overriden by precomputed value

### LINKED ISSUE #225: Formulas overwritten by the lazy value of the cell
It appears that formulas are overwritten by the default formula value in some cases. 

I looked a bit at the code and I found that when parsing the cell at the end of the `<v>` tag the parser doesn't check if the `<f>` element as already be encountered and just use the `Cell::set_value_string` type function which set the formula to `None`

I experimented with adding a check for the formula and used `set_formula_result_default` in case where a formula exists and my fix seems to work. But that said I'm not familiar with the codebase and it might break some other cases : 

```rust
// In src/structs/cell.rs

fn set_attributes</**/> (/**/) {
    // ...
    loop {
        match reader.read_event_into(&mut buf) {
            //...
            Ok(Event::End(ref e)) => match e.name().into_inner() {
                    b"v" => match type_value.as_str() {
                        _ if self.cell_value.is_formula() => {
                            self.set_formula_result_default(&string_value);
                        }
                        "str" => {
                            self.set_value_string(&string_value);
                        }
                        //...
                        _ => {}
                    },
                    // ...
                },
               //...
           }
     }
}
```

I've also seen the lazy value type and wonder if it could be related to this too.

oxabz:
Fixes #225 

## Change

Previously, if a cell had both the <f> and <v> tag it would delete the formula.
This pull request adds a check to the handling of the end event of the v tag to handle that scenario

## Question 

I'm not sure if the `Cell::set_formula_result_default` is the proper behaviour for that scenario

### DIFF
--- a/src/structs/cell.rs
+++ b/src/structs/cell.rs
@@ -329,8 +329,8 @@ impl Cell {
                 }
                 Ok(Event::End(ref e)) => match e.name().into_inner() {
                     b"v" => match type_value.as_str() {
-                        _ => if self.cell_value.is_formula() {
-                            self.set_value_lazy(&string_value);
+                        _ if self.is_formula() => {
+                            self.set_formula_result_default(&string_value);
                         }
                         "str" => {
                             self.set_value_string(&string_value);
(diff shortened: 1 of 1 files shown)


MathNya:
Thank you for the reference.

PR #229, opened 2024-09-23
fixup! Speed up image stuff (#223)

xamgore:
I didn't consider a few places where the functions could be used, so the version `2.1.0` [should be yanked](https://doc.rust-lang.org/cargo/commands/cargo-yank.html).

```toml
umya-spreadsheet = { version = "2.1.0", default-features = false }
```

Will fail.


MathNya:
@xamgore 
Thank you for contacting us.
We will respond as soon as possible.

Issue #231, opened 2024-10-05
Add Worksheet::copy_range for Copy/Paste cells?


piaoger:
I'd like to copy a row and insert a couple of rows with same values as below lines:
```rust
        let mut sheet1 = book.get_sheet_mut(&1).unwrap();
        sheet1.insert_new_row(&9, &3);
        copy_range(&mut sheet1, "A8:L8", &1, &0);
```

My copy_range is similar to Worksheet::move_range but for Copy/Paste workflow.Can we add a Worksheet::copy_range as below?
```rust
fn move_or_copy_range(..., remove: bool)
pub fn move_range();
pub fn copy_range();
```

MathNya:
@piaoger 
Thank you for contacting us.
Yes, I would like to know if you have any questions. I think it is probably possible.
We will address this in the next update.

PR #234, opened 2024-10-10
Add Worksheet::copy_range(#231)

piaoger:
Add Worksheet::copy_range Api to support Copy/Paste cells in worksheet.


MathNya:
@piaoger 
Thanks for the PR.
It is exactly the same implementation I had imagined.
There is some kind of error, but I will merge it.
(I will take care of the error.)

PR #236, opened 2024-10-25
Some changes

agentjill:
Found a bug in the `TrueFalseValue` struct's `get_value_string()` and other small changes.


MathNya:
@agentjill 
Thank you as always.
We will merge them.

PR #242, opened 2024-11-16
Reduce memory footprint by reducing size of data types

schungx:
This PR attempts to reduce the memory footprint of using this crate.

* `String` types are changed to `Box<str>` to save 8 bytes each.
* The `thin-vec` crate is used to replace `Vec`'s, saving 16 bytes each.
* Many large `Option`'s are `Box`ed
* Many large types in `HashMap`s are `Box`ed

I try to avoid making code-breaking changes, but some return types are changed, such as `&Vec<T>` to `&[T]`... in most cases that shouldn't matter.  The entire test suite compiles are run without change.

The current API design prevents more involved optimizations without breaking code.  For example, returning `&bool`, `&u32` etc. prevents those fields from being replaced by smaller, packed data types like `bitflags`. A future major version should really change those to return non-references instead.

I have given up on optimizing chart data types because they are simply too tedious, and there shouldn't really be more than a few charts in a typical spreadsheet.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,6 +29,7 @@ lazy_static = "1.4.0"
 md-5 = "0.10.6"
 regex = "1.11.1"
 sha2 = "0.10.8"
+thin-vec = "0.2.13"
 thousands = "0.2.0"
 quick-xml = { version = "0.36.2", features = ["serialize"] }
 zip = { version = "1.1.4", default-features = false, features = ["deflate"] }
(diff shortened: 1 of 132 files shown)


MathNya:
@schungx 
Thank you for the tremendous corrections.
We have no problem with the revisions and will approve them.

schungx:
Beware some function signatures have minor subtle differences which may require code changes. Be prepared for some incoming issues on this.

schungx:
When I have time I'll put in `#[inline]` on all the trivial APIs.

MathNya:
Noted.
We will include that in the release notes.

PR #243, opened 2024-11-19
Add #[inline] to short public functions

schungx:
This should help those users who don't build with LTO turned on.

### DIFF
--- a/src/helper/crypt.rs
+++ b/src/helper/crypt.rs
@@ -477,18 +477,21 @@ fn hash(algorithm: &str, buffers: Vec<&[u8]>) -> Result<Vec<u8>, String> {
     Ok(digest.finalize().to_vec())
 }
 
+#[inline]
 fn gen_random_16() -> Vec<u8> {
     let buf: &mut [u8] = &mut [0; 16];
     getrandom::getrandom(buf);
     buf.to_vec()
 }
 
+#[inline]
 fn gen_random_32() -> Vec<u8> {
     let buf: &mut [u8] = &mut [0; 32];
     getrandom::getrandom(buf);
     buf.to_vec()
 }
 
+#[inline]
 fn gen_random_64() -> Vec<u8> {
     let buf: &mut [u8] = &mut [0; 64];
     getrandom::getrandom(buf);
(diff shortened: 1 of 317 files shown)


src/helper/crypt.rs
```diff
@@ -477,25 +477,29 @@ fn hash(algorithm: &str, buffers: Vec<&[u8]>) -> Result<Vec<u8>, String> {
...
 
+#[inline]
 fn gen_random_64() -> Vec<u8> {
     let buf: &mut [u8] = &mut [0; 64];
     getrandom::getrandom(buf);
     buf.to_vec()
 }
 
 // Create a buffer of an integer encoded as a uint32le
+#[inline]
```
c-git, line 502: Are we sure all of these are more suitable to be inline? Did we check what impact this has on binary size?

schungx: Putting `#[inline]` on functions enable them to be optimized across crates.  This is a library crate, so if you don't have that and LTO is not on, the Rust compiler will never inline even very simple function such as getters.

Also, inlining usually enables more optimizations in the call site.  For example, the following code:
```rust
fn flip(x: bool) -> bool {
    if x { false } else { true }
}

// use it in another crate:
if flip(true) { ... }
```

If you don't compile with LTO, everything stays, including the call to `flip` because the Rust compiler will not inline across crate boundaries.  If you put `#[inline]` on `fn flip`, the entire code is optimized away to zero.

c-git: There is a cost in terms of compile time by doing that optimization. All I was saying is that by doing this we remove that choice from the end user. Whereas with LTO they can opt in if they want to.

schungx: AFAIK Rust doesn't inline at all in debug mode. That's why performance is way slower than release, because all those `as_ref` and `as_mut` etc are actually function calls where with inlining they turn to the original reference.

schungx: > Fourth, when building libraries, proactively add #[inline] to small non-generic functions. Pay special attention to impls: Deref, AsRef and the like often benefit from inlining. A library can’t anticipate all usages upfront, it makes sense to not prematurely pessimize future users. Note that #[inline] is not transitive: if a trivial public function calls a trivial private function, you need to #[inline] both. See this benchmark for details.

This section is actually my exact point. A library should proactively `#[inline]` wherever it makes sense.

MathNya:
@schungx
We will merge this information as well.
 I did not know the details about #[inline] either.
I have learned about the above comments as well.

schungx:
Just a follow up with observations.

I find that once using the new version with `#[inline]`y linking time is reduced to less than 1/3 of the time it took before.

I suspect that since Rust is now doing the inlining at HIR stage, it avoids having to do it with LTO. Looks like LTO takes a huge amount of time if there are many functions.

PR #247, opened 2024-12-08
Update rust edition to 2021

mxsrm:
Edition 2021 is stable for almost 3 years now. Edition 2024 will release soon. I think it's safe to upgrade the code to edition 2021.

~~I also changed the dependency versions in `Cargo.toml` to MAJOR.MINOR instead of MAJOR.MINOR.PATCH. This allows end-users to have security or urgent fixes of dependencies to be applied automatically without the need for this crate to release a new version every time.~~

Due to the change to rust edition 2021 the next release should probably be a MINOR release at least.

Thanks for your great project!


c-git:
I don't think the edition change affects user crates, I think at the cross crate level I think it's supposed to be the same rust.

I also don't think the patch version affects user crates other than enforcing a minimum. It doesn't constrain the maximum. Based on my understanding of Semver all newer versions with the same Major version would be Semver compatible and thus users are not blocked from updates within the same Major version.

mxsrm:
@c-git 
I checked the documentation, you are correct. I deleted the commit that changed the versions to MAJOR.MINOR.

MathNya:
@mxsrm 
Thanks for the PR.
We have no problem with the content, so we will merge it.

PR #249, opened 2024-12-15
Optimize code

mxsrm:
I enabled clippy lints and fixed all of them, making the code more idiomatic in many places.


mxsrm:
I added a [commit](https://github.com/MathNya/umya-spreadsheet/pull/249/commits/3cc7fbcf6d5b6459d86e68be3285731fd8161c30) to remove most of the &bool, &i32 etc. references.

This would make this PR breaking for the public API, so if you opt to include this commit, it should probably result in a major version bump.

schungx:
I suggest you open a separate PR on a new branch for the breaking changes, so people can cherry pick the ones they want.

mxsrm:
@schungx 
I can do that, but my breaking changes build upon the `clippy` fixes.

schungx:
You can open a new branch, say `vnext` with just the breaking changes, branching off the last existing `clippy` fixes commit.

MathNya:
@mxsrm 
Thanks for the PR.
We are planning to release Version 3.0.0 as our next update.
This will include some disruptive changes.
Since this will be a major fix, we have prepared a branch.
We will merge it here.

mxsrm:
@MathNya 
What is your timeline for 3.0.0? I probably have some time over the holidays to optimize memory consumption. I think we can replace many of the `Box<T>` constructs with more _direct_ implementations.

I really enjoy the Rust lang and this seems like a great project to learn a lot about the language. I figure we could probably remove all the wrapper Structs (like `Int16Value`) using generic traits for example. `Box<str>` could be replaced by `String`, or even better, `&str` where possible to avoid additional allocations etc.

schungx:
Well, `Box<str>` is two words, while `String` is three words.  You save 8 bytes by going with `Box<str>` if the string is never going to be modified, which seems to be the case with this API.  Values are usually replaced, not modified.

`&str` requires everything to have a lifetime so it will be major surgery.

`Box`-ing some stuff has the effect of reducing their footprint in case of `Option::None`.  For example, `Option<T>` where `T` is a large type will be as large as `T` or even larger, while `Option<Box<T>>` is only a single word regardless of how large `T` is.  Therefore, if that field contains `None` much of the time, it is a good candidate for redirection.

mxsrm:
@schungx 
It's debateable if saving a single `usize` per Item really changes anything performance wise, unless you develop for embeded systems.

Using `tinyvec` and alike you risk exposing yourself to thirdparty bugs, memory leaks etc. the `std::Vec` would not have.

schungx:
> @schungx 
> It's debateable if saving a single `usize` per Item really changes anything performance wise, unless you develop for embedded systems.
> 
> Using `ThinVec` and alike you risk exposing yourself to thirdparty bugs, memory leaks etc. the `std::Vec` would not have.

True, but in a large Excel spreadsheets, with such deep hierarchical data structures, you have a LOT of such small objects.

A struct with an array 99% empty multiplied by 1000 times is 16K just in additional overhead. And that is only one out of hundreds of settings possible for a cell. Think of a spreadsheet with 10,000 rows.

MathNya:
@mxsrm 
Thank you for your cooperation.
Version 3.0.0 accepts all fixes, including destructive fixes.
See here for specific milestones.
https://github.com/MathNya/umya-spreadsheet/milestone/2

Issue #248, opened 2024-12-12
Feature Request - Set Worksheet Gridlines


ice1080:
Hello!

Fantastic library, has been very helpful for our work. I was wondering if it would be possible to get the ability to set a worksheet's gridlines visibility. This would open up more abilities for us as we have some spreadsheets that have to have no gridlines, and setting every cell's background to white as a workaround isn't feasible for us due to size and performance reasons.

Thanks!

MathNya:
@ice1080 
The ability to hide Gridlines was not yet implemented.
This was a surprise to me as well.
This will be addressed in the next version.

ice1080:
@MathNya that is fantastic, thank you so much for the quick work. We'll test it out as soon as the new version is published.

mxsrm:
@ice1080 
You can test without waiting for a new release:

```Cargo.toml
[dependencies]
umya-spreadsheet = { git = "https://github.com/MathNya/umya-spreadsheet.git", rev = "5e200004fb0bec8b07a634524d9a8363ba0bc700" }
```

ice1080:
Thanks @mxsrm . My team tested with that today and it worked beautifully. Thanks again @MathNya !

MathNya:
@ice1080 
Sorry for the delay.
Version 2.2.1 has been released.
You can also check it here.

ice1080:
No problem at all. We've updated to v2.2.1 now

PR #250, opened 2024-12-22
Dependency Updates, Documentation Enhancements, New Features, and Code Refactoring

mxsrm:
So, I have the first big truck of changes for you. I mainly did formatting (those are separate commits) and clippy changes, improving alot of things efficiency-wise. I also fixed a MRSV of minimum supported Rust version, dictated to us by our dependencies anyway.

I compressed the `helper/crypt.rs` part and pulled away all the static values we currently do not allow to be changed from user side anyway. Makes the code alot more readable.

----

### Chore

 - <csr-id-ea1d8a281ac6492044a38e6787928207005e01d6/> update zip dependency to version 2.2.2
 - <csr-id-37d1139cd3588ae7ec6b89687e69b160e480d7e5/> add categories to Cargo.toml
 - <csr-id-de9f38ff8875ee2c9f6de0ededfaf48c2c3d3da4/> allow multiple versions of `thiserror`
   This fixes the following `clippy` warning:
   > multiple versions for dependency `thiserror`: 1.0.69, 2.0.8
   > for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#multiple_crate_versions
   
   `thiserror` and `thiserror-impl` are transitive dependencies:
   
   ```sh
   cargo tree --invert --package thiserror-impl@1.0.69
   
   thiserror-impl v1.0.69 (proc-macro)
   └── thiserror v1.0.69
       ├── html_parser v0.7.0
       │   └── umya-spreadsheet v2.2.0 (/home/mxsrm/code/umya-spreadsheet)
       └── rav1e v0.7.1
           └── ravif v0.11.11
               └── image v0.25.5
                   └── umya-spreadsheet v2.2.0 (/home/mxsrm/code/umya-spreadsheet)
   
   cargo tree --invert --package thiserror-impl@2.0.8
   
   thiserror-impl v2.0.8 (proc-macro)
   └── thiserror v2.0.8
       ├── pest v2.7.15
       │   ├── html_parser v0.7.0
       │   │   └── umya-spreadsheet v2.2.0 (/home/mxsrm/code/umya-spreadsheet)
       │   ├── pest_derive v2.7.15 (proc-macro)
       │   │   └── html_parser v0.7.0 (*)
       │   ├── pest_generator v2.7.15
       │   │   └── pest_derive v2.7.15 (proc-macro) (*)
       │   └── pest_meta v2.7.15
       │       └── pest_generator v2.7.15 (*)
       └── zip v2.2.2
           └── umya-spreadsheet v2.2.0 (/home/mxsrm/code/umya-spreadsheet)
   ```
 - <csr-id-995f40d6e0242e76c863c3f2f9b26645220442f9/> remove unused `js` feature from Cargo.toml
   This commit removes the unused `js` feature from the `[features]`
   section in `Cargo.toml`. The `js` feature is no longer utilized in
   the project, and its removal helps to maintain a cleaner configuration.
   Previously, this feature led to the inclusion of the `js` feature from
   the `getrandom` dependency, which was removed in a prior commit.
   
   - Deleted the `js` feature entry from `Cargo.toml`.
 - <csr-id-f76be7ab9ff98f40f2966cf06a3dfbebe24a0099/> set minimum supported Rust version to 1.79.0
   This commit sets the minimum supported Rust version in `Cargo.toml`
   to 1.79.0, as determined by the `cargo-msrv` tool. This version is
   dictated by the `bitstream-io` dependency, which is a transitive
   dependency of the `image` crate. Setting the minimum version ensures
   compatibility with the required features and functionality.
   
   - Added `rust-version = "1.79.0"` to `Cargo.toml`.
 - <csr-id-6b2c88ef972964c96154c024a213f98774e20838/> move lint configurations to lib.rs
   This commit removes the lint configurations from `Cargo.toml` and
   transfers them to `lib.rs` using attribute macros. This change
   centralizes lint settings within the codebase, making it easier to
   manage and understand the linting rules applied to the project.
   
   - Removed lint settings from `Cargo.toml`.
   - Added lint attributes in `lib.rs` for better backwards compatibility.
 - <csr-id-517cf8ca6e25adb6ee8ef44db1b6df0252b350b8/> sort dependencies in Cargo.toml
   This commit sorts the dependencies in `Cargo.toml` for improved
   readability and organization. Sorting helps maintain a consistent
   format and makes it easier to locate specific dependencies.

### Documentation

 - <csr-id-f8d9659e5dd7ef70b6386ebcf05972c5f214956f/> enhance documentation for encryption-related functions
   * Add detailed parameter descriptions and return values for all public functions
   * Improve documentation formatting and clarity for encryption methods
   * Add specific documentation for AES-256-CBC implementation details
   * Include clippy allow attributes for possible truncation warnings
   * Document HMAC and IV generation processes
   * Add comprehensive documentation for password-to-key conversion logic

### New Features

 - <csr-id-92a1af9cd1642e187432890b39851b7650cc4819/> enhance Excel timestamp conversion with additional calendar systems
   * add `num-traits` dependency for numeric casts
   * introduce `DEFAULT_TIMEZONE` constant for consistency
   * implement `excel_to_date_time_object` for converting Excel timestamps to `NaiveDateTime`
   * add `convert_date_windows_1900` and `convert_date_mac_1904` for specific calendar systems
   * enhance `convert_date_crate` to support both Windows 1900 and Mac 1904 systems
   * include detailed documentation for new functions with examples and panic behavior

### Refactor

 - <csr-id-4eab5d75a2ce9f661c7f8501b0fe80f1c10b93f7/> update is_address function for regex initialization
   - Refactored the `is_address` function to use `OnceLock` for lazy initialization of the regex pattern, improving performance by compiling the regex only once.
   - Enhanced documentation for `is_address` to clarify its parameters, return values, and potential panics.
   - Removed the previous implementation of `is_address` to streamline the code.
   - Added examples in the documentation to demonstrate usage.
 - <csr-id-505d76665bb58903fb7f01ca368641d95856cc30/> remove unnecessary qualification
 - <csr-id-8d41bc13cb1807b926c7c381e75baa35adeb287f/> clippy lints
 - <csr-id-72faf4f9f6672c0ae5ba2ca2c39dc4b701d22622/> simplify make_buffer function and improve worksheet processing
   - Reorganized the `make_buffer` function for better readability and maintainability.
   - Replaced manual iteration with `try_for_each` for processing worksheets.
   - Consolidated worksheet processing logic to handle both deserialized and raw data more cleanly.
   - Improved comments for clarity on each processing step.
   - Streamlined the addition of various worksheet components (charts, drawings, comments, etc.).
   - Ensured proper handling of printer settings and other worksheet relationships.
 - <csr-id-4dcea0eaa42901a528c8e8d5fb6ed99a45b189ab/> remove leading underscores from method names and allow dead_code at crate level
   - Rename private methods across multiple modules by removing leading underscores to adhere to Rust naming conventions.
   - Additionally, add `#![allow(dead_code)]` at the crate level to silence dead code warnings.
 - <csr-id-26e7021086c995c35b1ccb226b759a7e632405e9/> centralize constants into a dedicated module
   Extract multiple static variables into the `constants` module, eliminating the need to pass them as parameters in function calls. This refactor simplifies function signatures, centralizes configuration values, and enhances maintainability. Future updates may revert this approach to accommodate support for additional algorithms as the crate evolves.
 - <csr-id-b1916f722ef673a4e1cc1112864e12683090624b/> abstract random byte generation with macro and enhance RNG security
   - Introduced the `generate_random_bytes!` macro to eliminate repetitive code for generating random bytes.
   - Replaced `rand::thread_rng` with `rand::rngs::OsRng` to utilize a cryptographically secure random number generator.
   - Updated cryptographic key constants to use fixed-size arrays instead of slices for improved type safety.
   - Enhanced error handling in random byte generation by adding `expect` messages.
   - Removed unused imports and commented-out code to clean up the codebase.
   - Improved consistency in hash function implementations and key management.
   
   This refactoring improves code maintainability, readability, and security by centralizing random byte generation and ensuring the use of secure RNG sources.
 - <csr-id-9511418565d8ac1feaaf5a4d49f2ee6b7812371d/> replace `is_none_or` with match statements
   This commit refactors the code to replace the use of the `is_none_or`
   method with match statements. This change improves compatibility with
   a lower minimum supported Rust version, since `is_none_or` is only
   available in stable since Rust 1.82.0.
   
   - Updated `is_support` method in `TwoCellAnchor` to use match.
   - Refactored `_has_vertical` and `has_horizontal` methods in `MergeCells`
     to use match for start and end row/column checks.
 - <csr-id-739446e66d02e1cae93a7ed03c38c38491dab0e3/> replace hex decoding/encoding with hex-literal
   This commit simplifies the test code by replacing custom hex decoding
   and encoding functions with the `hex-literal` crate. This change
   improves readability and reduces boilerplate code in the tests.
   
   - Removed `decode_hex` and `encode_hex` helper functions.
   - Updated test cases to use the `hex!` macro for byte literals.
 - <csr-id-b65b0352a3e3353f847ae42760d17bb53e116fcb/> replace OsRng with thread_rng for random byte generation
   This commit refactors the random byte generation in the `gen_random_bytes` function by replacing the use of `OsRng` with `thread_rng`. This change simplifies the code and improves performance by using the thread-local random number generator.
   
   - Updated the `gen_random_bytes` function to utilize `thread_rng()` for filling the byte vector.
   - Removed the `OsRng` import as it is no longer needed.
 - <csr-id-12926d6cb8ea0f63826f0ca1f874ef548e759fbe/> replace ahash with std::hash::DefaultHasher
   This commit removes the `ahash` dependency and replaces its usage with the standard library's `DefaultHasher`. This change simplifies the dependency tree and reduces external dependencies while maintaining the same hashing functionality.
   
   - Removed the `ahash` crate from `Cargo.toml`.
   - Updated the `SharedStringItem` struct to use `std::hash::DefaultHasher` instead of `AHasher` from `ahash`.
 - <csr-id-cb80120e9be70463eb31c175648eb3d0ef6bd129/> rename cell_collection to cells, row_dimensions to rows, and column_dimensions to columns
   - Renamed `cell_collection` to `cells` for consistency and clarity.
   - Renamed `row_dimensions` to `rows` for improved readability.
   - Renamed `column_dimensions` to `columns` for better understanding.
   - Updated all method calls and references accordingly.
 - <csr-id-714bf078ad985a82c9063f6ff558ab6d40b5ef1d/> replace `ThinVec` with rust-lang `Vec`
 - <csr-id-aaaec5f799546a2eba7e70e526fb92d062c9f914/> replace `lazy_static` with built-in `std::sync::OnceLock`
 - <csr-id-f3de7699e001e338867b80dc09ce454d54d1a858/> remove lazy_static, add comments
 - <csr-id-71f0572d6e2cb95a2d1d840880df66967727f877/> fix typos, clippy lints
 - <csr-id-30e3e9a7df0a8ce56e1b39c7ab879f2708376897/> reimplement and cleanup `helper/crypt.rs`
 - <csr-id-a317cceed76af2f55178237aafb89b0bc0229048/> replace `getrandom` with `rand` for random byte generation
   - Replaced `getrandom` crate with `rand` to simplify and modernize random number generation.
   - Updated `gen_random_*` functions to use `rand::thread_rng().fill()` for generating random bytes of specified lengths.
   - Adjusted encryption functions to use `gen_random_bytes` for generating random values.

### Style

 - <csr-id-e60fa408723c89b87c52275080ecefe6081cf30d/> fix many clippy lints
 - <csr-id-6b3c774b8b53bb64b910f918c6ab21ad442d1929/> fix formatting and improve check_sheet_name logic
   - Corrected the documentation comment for `get_sheet_by_name` to fix the syntax error.
   - Refactored the `check_sheet_name` method to use a more concise if-else structure for clarity.
   - Improved readability of the code while maintaining the same functionality.
 - <csr-id-98fcf7f9adadf415e8776db5d58b2297b8b53201/> add modern rustfmt.toml and apply it
 - <csr-id-48b864bc3217b84586cdb7787819001118e53e8d/> enable more lints

### Commit Statistics

<csr-read-only-do-not-edit/>

 - 33 commits contributed to the release.
 - 11 days passed between releases.
 - 31 commits were understood as [conventional](https://www.conventionalcommits.org).
 - 1 unique issue was worked on: [#249](https://github.com/mxsrm/umya-spreadsheet/issues/249)

### Commit Details

<csr-read-only-do-not-edit/>

<details><summary>view details</summary>

 * **[#249](https://github.com/mxsrm/umya-spreadsheet/issues/249)**
    - Optimize code ([`1f81738`](https://github.com/mxsrm/umya-spreadsheet/commit/1f81738f446e13651bb1b1e05ee0177655fd5878))
 * **Uncategorized**
    - Fix many clippy lints ([`e60fa40`](https://github.com/mxsrm/umya-spreadsheet/commit/e60fa408723c89b87c52275080ecefe6081cf30d))
    - Fix formatting and improve check_sheet_name logic ([`6b3c774`](https://github.com/mxsrm/umya-spreadsheet/commit/6b3c774b8b53bb64b910f918c6ab21ad442d1929))
    - Update zip dependency to version 2.2.2 ([`ea1d8a2`](https://github.com/mxsrm/umya-spreadsheet/commit/ea1d8a281ac6492044a38e6787928207005e01d6))
    - Update is_address function for regex initialization ([`4eab5d7`](https://github.com/mxsrm/umya-spreadsheet/commit/4eab5d75a2ce9f661c7f8501b0fe80f1c10b93f7))
    - Remove unnecessary qualification ([`505d766`](https://github.com/mxsrm/umya-spreadsheet/commit/505d76665bb58903fb7f01ca368641d95856cc30))
    - Add categories to Cargo.toml ([`37d1139`](https://github.com/mxsrm/umya-spreadsheet/commit/37d1139cd3588ae7ec6b89687e69b160e480d7e5))
    - Allow multiple versions of `thiserror` ([`de9f38f`](https://github.com/mxsrm/umya-spreadsheet/commit/de9f38ff8875ee2c9f6de0ededfaf48c2c3d3da4))
    - Clippy lints ([`8d41bc1`](https://github.com/mxsrm/umya-spreadsheet/commit/8d41bc13cb1807b926c7c381e75baa35adeb287f))
    - Enhance documentation for encryption-related functions ([`f8d9659`](https://github.com/mxsrm/umya-spreadsheet/commit/f8d9659e5dd7ef70b6386ebcf05972c5f214956f))
    - Simplify make_buffer function and improve worksheet processing ([`72faf4f`](https://github.com/mxsrm/umya-spreadsheet/commit/72faf4f9f6672c0ae5ba2ca2c39dc4b701d22622))
    - Remove leading underscores from method names and allow dead_code at crate level ([`4dcea0e`](https://github.com/mxsrm/umya-spreadsheet/commit/4dcea0eaa42901a528c8e8d5fb6ed99a45b189ab))
    - Centralize constants into a dedicated module ([`26e7021`](https://github.com/mxsrm/umya-spreadsheet/commit/26e7021086c995c35b1ccb226b759a7e632405e9))
    - Add modern rustfmt.toml and apply it ([`98fcf7f`](https://github.com/mxsrm/umya-spreadsheet/commit/98fcf7f9adadf415e8776db5d58b2297b8b53201))
    - Abstract random byte generation with macro and enhance RNG security ([`b1916f7`](https://github.com/mxsrm/umya-spreadsheet/commit/b1916f722ef673a4e1cc1112864e12683090624b))
    - Enhance Excel timestamp conversion with additional calendar systems ([`92a1af9`](https://github.com/mxsrm/umya-spreadsheet/commit/92a1af9cd1642e187432890b39851b7650cc4819))
    - Remove unused `js` feature from Cargo.toml ([`995f40d`](https://github.com/mxsrm/umya-spreadsheet/commit/995f40d6e0242e76c863c3f2f9b26645220442f9))
    - Set minimum supported Rust version to 1.79.0 ([`f76be7a`](https://github.com/mxsrm/umya-spreadsheet/commit/f76be7ab9ff98f40f2966cf06a3dfbebe24a0099))
    - Replace `is_none_or` with match statements ([`9511418`](https://github.com/mxsrm/umya-spreadsheet/commit/9511418565d8ac1feaaf5a4d49f2ee6b7812371d))
    - Move lint configurations to lib.rs ([`6b2c88e`](https://github.com/mxsrm/umya-spreadsheet/commit/6b2c88ef972964c96154c024a213f98774e20838))
    - Sort dependencies in Cargo.toml ([`517cf8c`](https://github.com/mxsrm/umya-spreadsheet/commit/517cf8ca6e25adb6ee8ef44db1b6df0252b350b8))
    - Replace hex decoding/encoding with hex-literal ([`739446e`](https://github.com/mxsrm/umya-spreadsheet/commit/739446e66d02e1cae93a7ed03c38c38491dab0e3))
    - Replace OsRng with thread_rng for random byte generation ([`b65b035`](https://github.com/mxsrm/umya-spreadsheet/commit/b65b0352a3e3353f847ae42760d17bb53e116fcb))
    - Replace ahash with std::hash::DefaultHasher ([`12926d6`](https://github.com/mxsrm/umya-spreadsheet/commit/12926d6cb8ea0f63826f0ca1f874ef548e759fbe))
    - Rename cell_collection to cells, row_dimensions to rows, and column_dimensions to columns ([`cb80120`](https://github.com/mxsrm/umya-spreadsheet/commit/cb80120e9be70463eb31c175648eb3d0ef6bd129))
    - Replace `ThinVec` with rust-lang `Vec` ([`714bf07`](https://github.com/mxsrm/umya-spreadsheet/commit/714bf078ad985a82c9063f6ff558ab6d40b5ef1d))
    - Replace `lazy_static` with built-in `std::sync::OnceLock` ([`aaaec5f`](https://github.com/mxsrm/umya-spreadsheet/commit/aaaec5f799546a2eba7e70e526fb92d062c9f914))
    - Remove lazy_static, add comments ([`f3de769`](https://github.com/mxsrm/umya-spreadsheet/commit/f3de7699e001e338867b80dc09ce454d54d1a858))
    - Fix typos, clippy lints ([`71f0572`](https://github.com/mxsrm/umya-spreadsheet/commit/71f0572d6e2cb95a2d1d840880df66967727f877))
    - Reimplement and cleanup `helper/crypt.rs` ([`30e3e9a`](https://github.com/mxsrm/umya-spreadsheet/commit/30e3e9a7df0a8ce56e1b39c7ab879f2708376897))
    - Replace `getrandom` with `rand` for random byte generation ([`a317cce`](https://github.com/mxsrm/umya-spreadsheet/commit/a317cceed76af2f55178237aafb89b0bc0229048))
    - Enable more lints ([`48b864b`](https://github.com/mxsrm/umya-spreadsheet/commit/48b864bc3217b84586cdb7787819001118e53e8d))
    - #248 ([`5e20000`](https://github.com/mxsrm/umya-spreadsheet/commit/5e200004fb0bec8b07a634524d9a8363ba0bc700))
</details>


schungx:
Two suggestions:

* I would suggest leaving `ThinVec` alone unless it is absolutely ok to add 16 bytes to each of those structures.

* `ahash`'s hasher is much faster than Rust's default hasher. Therefore `ahash` is usually used when performance is concerned.  I suggest doing some profiling to check.

These two changes may affect CPU and RAM usage on low level hardware such as ARM chips or low power x86.

mxsrm:
`ahash` was only used in a single place:

https://github.com/MathNya/umya-spreadsheet/blob/b58e16edb0bb11cca9561dd7db898183be551002/src/structs/shared_string_item.rs#L68-L81

All other `Hash`s were already using the default implementation. Do you really see that much difference in performance?

Regarding `ThinVec`: have you measured your use cases against this PR? Rust does a lot of clever optimization under the hood and `ThinVec` even tells you, that if you construct it, the allocations might be actually more than `std::Vec`:

> Of course, any actually constructed ThinVec will theoretically have a bigger allocation, but the fuzzy nature of allocators means that might not actually be the case.

mxsrm:
One big area one could think about improving would be error handling. Currently (this PR) we have 601 `.unwrap()` calls in this codebase.

src/helper/crypt/key.rs
```diff
@@ -0,0 +1,199 @@
...
+/// * `password` - The password string to hash
+/// * `salt` - The salt bytes to use in hashing
+/// * `spin_count` - Number of iterations for the hash function
+///
+/// # Returns
+/// A vector containing the final hash value
+#[allow(clippy::cast_possible_truncation)]
+pub(crate) fn convert_password_to_hash(password: &str, salt: &[u8], spin_count: usize) -> Vec<u8> {
+    // Convert password to UTF-16LE bytes
+    let password_bytes: Vec<u8> = password.encode_utf16().flat_map(u16::to_le_bytes).collect();
```
mxsrm, line 186: The old implementation was more complicated, but this yields the same results:

```rust
fn main() {
    let pw = "abcdefg";
    
    let a = {
        let mut password_buffer: Vec<u8> = Vec::new();
        let v: Vec<u16> = pw.encode_utf16().collect();
        for a in v {
            let d = a.to_le_bytes();
            password_buffer.push(d[0]);
            password_buffer.push(d[1]);
        }
        password_buffer
    };
    
    let b: Vec<u8> = pw.encode_utf16().flat_map(u16::to_le_bytes).collect();
    
    assert_eq!(a, b);
    
    println!("{:?}", a); // [97, 0, 98, 0, 99, 0, 100, 0, 101, 0, 102, 0, 103, 0]
    println!("{:?}", b); // [97, 0, 98, 0, 99, 0, 100, 0, 101, 0, 102, 0, 103, 0]
}
```

[playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b57143735ddf33477fd2060f50f017cc)

schungx:
> Regarding `ThinVec`: have you measured your use cases against this PR? Rust does a lot of clever optimization under the hood and `ThinVec` even tells you, that if you construct it, the allocations might be actually more than `std::Vec`:

Well, Rust layout follows a very stable pattern and it it one of the things that Rust does not optimize away for you.  Therefore a large type stays a large type after all optimizations.  That's why you still need to consider size of types in Rust programs.  AFAIK Rust never touches the memory layout of types.

`ThinVec` is used in the FireFox browser pretty extensively so I suppose it would be stable enough.

An alternative is to use `Option<Box<Vec>>` and automatically puts in code to change empty `Vec`'s into `None`. This has the same effect as using `ThinVec`.  Using `Box<Vec>` itself is also a solution, which also has the same size as using `ThinVec`, but then you have an extra allocation for each instance.  Memory usage does not change, but you pack the parent type more closely so you get better cache locality.

Also remember collections such as `HashMap` may contain up to 70% empty slots, meaning that if your type contains empty spaces then those empty slots just multiply.

mxsrm:
> `ThinVec` is used in the FireFox browser pretty extensively so I suppose it would be stable enough.

Firefox uses [smallvec](https://crates.io/crates/smallvec) which is developed (in part) by [Mozilla](https://github.com/servo/rust-smallvec) and is actively maintained. It has a different architecture and idea behind it then the (seemling unmaintained) [thin-vec](https://crates.io/crates/thin-vec) you refer to.

schungx:
> Currently this crate mostly exists to facilitate Gecko (Firefox) FFI, but it works perfectly fine as a native rust library as well.

Not sure myself but it says it is developed to interop with Gecko.

Anyway, I'm not particularly hung up on it. However I have maintained crates that using it makes a lot of sense, especially where there is an enum with a `Vec` that makes it 16 bytes larger and it kills caching. Can't move to `Box<Vec>` because it forces another allocation and redirection that slows things down.  In such cases `ThinVec` has a clear benefits in tight loop benchmarks.

mxsrm:
I'm happy to reimplement it if @MathNya thinks it useful. Or implement `smallvec` instead.

schungx:
> I'm happy to reimplement it if @MathNya thinks it useful. Or implement `smallvec` instead.

`SmallVec` solves a different problem: avoiding allocations. It does not save memory.  A `SmallVec` can be larger than a `Vec` depending on your config.

You'd need to bemchmark to check if using `SmallVec` is worth it as it stores small number of items inline to avoid allocations, at the expense of larger memory footprint.

schungx:
A bit more details on the two different crates.

`SmallVec` I suggest we use on transitory purposes, for example preparing a list for processing inside a function.  The list is destroyed after use and never goes outside of the function.  Using `SmallVec` this way keeps everything on the stack for direct access and cache locality, and avoids all allocations if the number of items is often small.  I have benchmarked a lot of such usages and it usually yields 10-15% speed improvements alone.

`ThinVec` should be used where every byte counts.  If using a `Vec` does not affect a type's size (e.g. when another variant is larger) or the `Vec` is usually empty (where you can use `Box<Vec>`), then there is no need.  I tend to use it in cases where having a `Vec` automatically makes a LOT of types larger, for example in a fundamental type which is embedded everywhere inside a lot of other data structures.

mxsrm:
If you want to substantially change the API you would end in a near-total re-write of the whole crate. If you do that, you could implement many things more efficiently, e.g. you could abstract the whole Unit types into a single generic one using `Cow<'_, str>`, e.g.:

```rust
use std::{
    borrow::Cow,
    fmt::Display,
    str::FromStr,
};

#[derive(Clone, Default, Debug, PartialEq, Eq, PartialOrd, Ord)]
pub struct Value<T> {
    value: Option<T>,
}

impl<T> Value<T> {
    pub(crate) fn get_value(&self) -> Option<&T> {
        self.value.as_ref()
    }

    pub(crate) fn set_value(&mut self, value: T) -> &mut Self {
        self.value = Some(value);
        self
    }

    pub(crate) fn remove_value(&mut self) -> &mut Self {
        self.value = None;
        self
    }

    pub(crate) fn has_value(&self) -> bool {
        self.value.is_some()
    }
}

impl<T: Display + Default + Clone + FromStr> Value<T> {
    pub(crate) fn set_value_string<S: Into<String>>(&mut self, value: S) -> &mut Self {
        match value.into().parse::<T>() {
            Ok(parsed) => self.set_value(parsed),
            Err(_) => self.remove_value(), // Or handle the error differently, e.g., log it
        }
    }

    pub(crate) fn get_value_or_default(&self) -> T {
        self.value.clone().unwrap_or_default()
    }
}

impl<'a> Value<Cow<'a, str>> {
    pub(crate) fn get_value_string(&self) -> Cow<'a, str> {
        self.value.clone().unwrap()
    }
}

macro_rules! create_and_export_ValueType {
    ($t:ty, $i:ident) => {
        pub type $i = Value<$t>;

        impl $i {
            pub(crate) fn get_value_string(&self) -> Cow<'a, str> {
                self.value
                    .as_ref()
                    .map_or_else(|| Cow::Borrowed(""), |v| Cow::Owned(v.to_string()))
            }
        }
    };
}

create_and_export_ValueType!(bool, BoolValue);
create_and_export_ValueType!(u8, ByteValue);
create_and_export_ValueType!(i8, SignedByteValue);
create_and_export_ValueType!(f64, DoubleValue);
create_and_export_ValueType!(i16, I16Value);
create_and_export_ValueType!(i32, I32Value);
create_and_export_ValueType!(i64, I64Value);
create_and_export_ValueType!(u16, U16Value);
create_and_export_ValueType!(u32, U32Value);
create_and_export_ValueType!(u64, U64Value);
pub type StringValue<'a> = Value<Cow<'a, str>>;
```

I started on that adventure once or twice, but you just need to change *too* much to do it incrementally. Many functions want `&str` from the `get_value_string` method, others expect `String` etc.

schungx:
Agree. Not recommended to change the current API too much as it'll be too much rewrites for existing code.

mxsrm:
In terms of memory efficiency (@schungx) I did some profiling with valgrind and the memory consumption upon cloning a Worksheet is huge because of the way `Color`, `Border` and `Borders` are represented. This is `v2.2.1`:

master_border
master_borders
master_color

This is the memory layout after re-implementing the structs, using the `rgb` crate as basis to represent colors:

pr_border
pr_borders
pr_color

I personally use umya to parse a plan from work, iterate over it, and color cells that contain specific names (to create personalized versions for many people at once).

Here massif data from Valgrind:

## Master (v2.2.1)

image

## This PR @ https://github.com/MathNya/umya-spreadsheet/pull/250/commits/89016c76ca92580bba3f0a27a7375239204f1054

image

## This PR @ [3c26246](https://github.com/MathNya/umya-spreadsheet/pull/250/commits/3c26246b66a0e13a12012b14a3bc7bdd0fee0d0c)

image

-----

We could probably do alot more optimization, but this change was minimal-invasive and did not require changing alot of internal APIs.

mxsrm:
It is interesting to see what happens if you remove all the `#[inline]` instructions and let the compiler decide for himself (otherwise same code as this PR @ [3c26246](https://github.com/MathNya/umya-spreadsheet/pull/250/commits/3c26246b66a0e13a12012b14a3bc7bdd0fee0d0c)).

### No `#[inline]`, lto = false

image

### No `#[inline]`, lto = true

image

So letting the compiler decide for itself and using lto if possible in the final application is better than force-inlining that many functions. I will push a commit to this PR to remove all `#[inline]` instructions.

src/structs/color.rs
```diff
@@ -1,111 +1,518 @@
...
+        } else if hex.len() == 6 {
+            let a = 0xFF;
+            let r = u8::from_str_radix(&hex[0..2], 16).ok()?;
+            let g = u8::from_str_radix(&hex[2..4], 16).ok()?;
+            let b = u8::from_str_radix(&hex[4..6], 16).ok()?;
+
+            Some(ARGB8 { a, r, g, b })
+        } else if hex.len() < 6 {
+            // To pass an integration test where the hex string is "#333".
+            // https://github.com/MathNya/umya-spreadsheet/pull/113
```
mxsrm, line 483: I am not sure if this should be valid behaviour? Is `#333` really a valid color? I added padding so the previous integration test added in PR #113 still passes.

mxsrm: This is the code.

MathNya: @mxsrm 
Hmmm, this seems to be a mistake for #333333.

MathNya:
@schungx 
@mxsrm 
Thank you for your support through the end of the year.
I am not very knowledgeable about performance enhancement and have left it to you.
I would be grateful if you could continue to support us if you would like.
I look forward to working with you again next year.

c-git:
I don't know if it's valid there but in CSS it's allowed. #333 is short for #333333

As per: https://www.w3schools.com/css/css_colors_hex.asp

> # 3 Digit HEX Value
> Sometimes you will see a 3-digit hex code in the CSS source.
> The 3-digit hex code is a shorthand for some 6-digit hex codes.
> The 3-digit hex code has the following form:
> #rgb
> Where r, g, and b represent the red, green, and blue components with values between 0 and f.

schungx:
>I will push a commit to this PR to remove all `#[inline]` instructions.

I would not suggest that.  My own benchmarks have `#[inline]` substantially speeding up the linking phase when LTO is turned on.

If LTO is turned off, any function without `#[inline]` will never be inlined, and there are a _LOT_ of simple accessor functions.

`#[inline]` is never for memory, it is intended for speed.  Without `#[inline]`, my release build links in 4 hours with LTO, while with `#[inline]` is it less than 1 hour.

mxsrm:
> I don't know if it's valid there but in CSS it's allowed. #333 is short for #333333
> 
> As per: https://www.w3schools.com/css/css_colors_hex.asp
> 
> > # 3 Digit HEX Value
> > Sometimes you will see a 3-digit hex code in the CSS source.
> > The 3-digit hex code is a shorthand for some 6-digit hex codes.
> > The 3-digit hex code has the following form:
> > #rgb
> > Where r, g, and b represent the red, green, and blue components with values between 0 and f.

I implented it in that way. See a496cbe.

mxsrm:
> @schungx @mxsrm Thank you for your support through the end of the year. I am not very knowledgeable about performance enhancement and have left it to you.

Not my area of expertise either. In daily life I am an interventional neuroradiologist that enjoys coding in his free time. @schungx is more knowledgable in that regard I think.

schungx:
> Not my area of expertise either. In daily life I am an interventional neuroradiologist that enjoys coding in his free time. @schungx is more knowledgable in that regard I think.

Just different experiences that's all.

I happen to be maintaining an LOB application running on limitrd hardware, so memory and CPU count a lot.  For most applications I'd agree there may not be a big impact, unless we're dealing with massive spreadsheets with tens of thousands of rows, like some users who complained in issues.

MathNya:
@mxsrm 
Thank you for your response.
We will merge them.

mxsrm:
> @mxsrm Thank you for your response. We will merge them.

Consider re-doing the merge without squashing into a single commit. If in the future you have an obscure bug and need to bisecet it will be near impossible to do so with such a giant commit.

The biggest parts of changes were rustfmt operations that do not change progam behaviour.

MathNya:
@mxsrm 
Yes, sir!

PR #254, opened 2025-01-17
small changes

agentjill:
these are small changes to reduce code size. May be merged if seems ok


MathNya:
@agentjill 
Thank you very much.
No problem, will merge.

PR #262, opened 2025-02-06
Fix copy col/row styling

schungx:
The current functions to copy col/row styling neglect to copy the column-wide/row-wide styles.


MathNya:
@schungx 
Thanks for the PR.
We will merge them.

Issue #263, opened 2025-02-06
Bug: get_highest_row returns column count and get_highest_col returns row count after updating to 2.2.2


DaimoniX:
![Image](https://github.com/user-attachments/assets/e416f65f-f1a5-4fd9-b7a1-6c3eb7f09e0c)

First image is output of 2.2.1, second output of 2.2.2
![Image](https://github.com/user-attachments/assets/5e9ef5d4-93fe-4402-af2f-43d21d14f9d0)

Likely caused by #256

PR #264, opened 2025-02-06
Fix flipped get_highest_column_and_row values

DaimoniX:
Addresses #263


MathNya:
@DaimoniX
Thanks for the PR.
We have confirmed that the modifications are correct and will merge them.
This is a very serious bug and we will release 2.2.3 as soon as possible.

PR #270, opened 2025-03-04
changes for optimisation

agentjill:
This PR contains following:
- Removed unnecessary data cloning into as `String` where the primitives can be used.
- Removed call to `.as_str()` where reference is suffice.
- Changed function signature for  some `pub(crate)` functions to accept primitive types.
- Simplified functions by removing unnecessary getters where `&mut Self` is passed into the function.
- Other minor improvements


MathNya:
@agentjill 
Thank you very much.
Merged.
If possible, please PR to version 3.0 for any future modifications regarding optimization.
https://github.com/MathNya/umya-spreadsheet/tree/version3.0

PR #273, opened 2025-03-15
Add indices to Cells.

schungx:
This PR adds two indices to the `Cells` type: one sorted by row then by column, the other sorted by column then by row.

Iterators are provided for different styles of sorting.

It is now extremely easy to iterate through all the cells of a particular row or column, and to iterate through cells in order.

Original methods that create `Vec`'s are deprecated in favor of using iterators as much as possible.

### DIFF
--- a/src/structs/cells.rs
+++ b/src/structs/cells.rs
@@ -34,6 +34,22 @@ impl Cells {
         self.iter_collection().collect()
     }
 
+    #[inline]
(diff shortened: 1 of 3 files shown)


c-git:
I didn't finish all but it looks good

MathNya:
@schungx 
Thanks for the PR.
I have confirmed that there is no problem and will merge it.
Version 2.2.4 will be released soon.

PR #287, opened 2025-07-05
Preserve Window Location, Size and tab ratio when editing an existing document

yerlibilgin:
When an existing document is edited and saved as a new file, its location, size and tab ratio are not preserved. This causes the new excel file window size to be small, and forces the user to resize it. This commit reads the values from the existing file, and uses them when saving the document.


MathNya:
@yerlibilgin 
Thanks for the PR.
We will merge them.

PR #308, opened 2026-02-13
fix: unescape hyperlink attrs + parse unprefixed drawing tags

wolfiesch:
## Summary
- Fix hyperlink parsing by XML-unescaping `location`, `tooltip`, and relationship targets.
- Improve drawing XML parsing to accept both `xdr:*`-prefixed tags and unprefixed tags (e.g. `wsDr`, `oneCellAnchor`, `twoCellAnchor`, marker children), matching files produced by other writers like OpenPyXL.

## Why
- ExcelBench Tier 2 fixtures include hyperlinks with escaped query parameters and drawings that use default namespaces without `xdr:` prefixes. Without these changes, hyperlink targets differ (`&amp;` vs `&`) and images/drawings fail to load.

## Tests
- `cargo test`

### DIFF
--- a/src/reader/xlsx/drawing.rs
+++ b/src/reader/xlsx/drawing.rs
@@ -27,7 +27,7 @@ pub(crate) fn read(
     xml_read_loop!(
         reader,
         Event::Start(ref e) => {
-            if e.name().into_inner() == b"xdr:wsDr" {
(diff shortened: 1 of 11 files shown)


MathNya:
@wolfiesch 
Thank you for the PR.
There are no issues with the changes, so I will proceed with the merge.

There is an error in the linter, but I will handle it.

PR #309, opened 2026-02-17
Fix formula cached-value typing and add typed formula result setters

PSU3D0:
## Summary

This PR improves formula cached-value writeback correctness and adds explicit APIs for typed cached results.

It is intentionally split into two commits:

1. **fix(cell): serialize formula cache with correct cell types**
2. **feat(cell): add typed formula cached-result setters**

## Motivation

This work is grounded in a concrete downstream integration need from **Formualizer**.

Formualizer recently added Rust-native workbook recalculation/writeback (`recalculate_file`) using `umya-spreadsheet` as the writer path. During validation, formula cached values written back to XLSX were observed as string-typed payloads in XML (`t="str"`) even when the computed result was numeric, boolean, or error.

That behavior creates type drift in `sheet*.xml` and can degrade interoperability with Excel/OpenXML consumers that rely on correct cached value typing.

Additionally, Formualizer needed an explicit, type-safe way to set formula cached results (number/bool/error/string/blank) while preserving formula metadata.

While this was identified through Formualizer, the fix is broadly useful: it improves formula-cache correctness for all `umya-spreadsheet` users.

### 1) Typed formula cached-value serialization

- Formula cells now emit type according to cached raw value kind:
  - numeric -> numeric path (no forced `t="str"`)
  - bool -> `t="b"` + `1/0`
  - error -> `t="e"` + actual error token
  - string -> `t="str"`
  - blank -> empty value
- Error serialization now writes the **actual cached error value** (e.g. `#N/A`) instead of hardcoding `#VALUE!`.

### 2) New typed API helpers (backwards-compatible)

Added on both `CellValue` and `Cell`:

- `set_formula_result_number(...)`
- `set_formula_result_bool(...)`
- `set_formula_result_error(...)`
- `set_formula_result_string(...)`
- `set_formula_result_blank()`

These methods update cached value (`<v>`) **without removing formula metadata** (`<f>`).

Existing `set_formula_result_default(...)` remains unchanged.

## Tests

Added regression coverage for:

- formula cached-value type mapping
- typed XML output for number/bool/error/string/blank formula caches
- roundtrip read behavior
- shared-formula metadata stability after write
- typed helper behavior preserving formula text

### Commands run

- `cargo test` (full suite) ✅
- targeted tests for new scenarios ✅

> Note: clippy `-D warnings` currently has pre-existing baseline failures on upstream `master` unrelated to this PR. This PR does not introduce new clippy debt in touched logic.

## Backwards compatibility

- No breaking API changes.
- Existing methods and call patterns continue to work.
- Output changes are correctness-oriented (typed cache serialization) while preserving formula objects/metadata.

## Why this helps

- Fixes a real production integration path in **Formualizer** (single-call workbook recalc + cache writeback).
- Better Excel/OpenXML interoperability via correctly typed formula cached values.
- Cleaner writeback semantics for any engine that computes and persists formula caches.
- More explicit and safer API for setting cached results programmatically.

### DIFF
--- a/src/structs/cell.rs
+++ b/src/structs/cell.rs
@@ -351,6 +351,39 @@ impl Cell {
     }
 
     #[inline]
+    pub fn set_formula_result_number<T>(&mut self, value: T) -> &mut Self
+    where
+        T: Into<f64>,
+    {
+        self.cell_value.set_formula_result_number(value);
+        self
+    }
+
+    #[inline]
+    pub fn set_formula_result_bool(&mut self, value: bool) -> &mut Self {
+        self.cell_value.set_formula_result_bool(value);
+        self
+    }
+
+    #[inline]
+    pub fn set_formula_result_error(&mut self, value: crate::CellErrorType) -> &mut Self {
+        self.cell_value.set_formula_result_error(value);
+        self
+    }
+
+    #[inline]
+    pub fn set_formula_result_string<S: Into<String>>(&mut self, value: S) -> &mut Self {
+        self.cell_value.set_formula_result_string(value);
+        self
+    }
+
+    #[inline]
+    pub fn set_formula_result_blank(&mut self) -> &mut Self {
+        self.cell_value.set_formula_result_blank();
+        self
+    }
+
+    #[inline]
     pub fn set_blank(&mut self) -> &mut Self {
         self.cell_value.set_blank();
         self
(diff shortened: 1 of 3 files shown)


MathNya:
@PSU3D0 
Thank you for your PR.
There are no issues with the content, so I will proceed with the merge.

Issue #310, opened 2026-03-01
Multiple unwrap() panics on valid XLSX files from LibreOffice and Apache POI


developer0hye:
## Summary

We're using `umya-spreadsheet` (v2.3.3) in [office2pdf](https://github.com/developer0hye/office2pdf) to parse XLSX files. When processing a corpus of ~2,800 real-world XLSX files (sourced from LibreOffice and Apache POI test suites), 15 files cause panics due to `unwrap()` calls in the library. We've worked around this with `catch_unwind`, but the panics should ideally be handled gracefully.

Related to #271 (panic reduction proposal) and #301 (FileNotFound on .xlsm).

### FileNotFound panics (7 files)
`unwrap()` on missing zip entries for hyperlinks/charts:
- `chart_hyperlink.xlsx`, `hyperlink.xlsx`, `tdf130959.xlsx`, `tdf115192.xlsx` (LibreOffice)
- `47504.xlsx`, `bug63189.xlsx`, `ConditionalFormattingSamples.xlsx` (Apache POI)

Stack trace points to zip entry lookup by relationship ID where the referenced file doesn't exist in the archive (e.g., external hyperlinks).

### ParseFloatError / ParseIntError panics (3 files)
`unwrap()` on `str::parse::<f64>()` or `str::parse::<u32>()`:
- `check-boolean.xlsx` — Boolean cell value parsed as float
- `functions-excel-2010.xlsx` — Integer overflow parsing cell reference
- `FormulaEvalTestData_Copy.xlsx` — Integer overflow parsing cell reference

### unwrap() on None panics (3 files)
Expected XML elements/attributes missing:
- `tdf100709.xlsx`, `64450.xlsx`, `sample-beta.xlsx`

### dataBar end element panics (2 files)
Conditional formatting parser panics via `panic!()`:
- `tdf162948.xlsx`, `NewStyleConditionalFormattings.xlsx`

Error: `Could not find dataBar end element`

## Expected behavior

All of these should return `Err` instead of panicking. Library consumers should be able to handle failures without `catch_unwind`.

## Environment
- `umya-spreadsheet` v2.3.3
- `office2pdf` v0.3.0
- Rust 1.85+
- Files sourced from LibreOffice and Apache POI test suites (publicly available)

Issue #271, opened 2025-03-08
Proposal: Gracefully handling errors and panic reduction in the library.


agentjill:
@MathNya, For [version3.0](https://github.com/MathNya/umya-spreadsheet/tree/version3.0), I am proposing following

1. Functions like (there are multiple, given only one as an example) `get_sheet` returns `Option<&Worksheet>`. The functions gives `None` when it cant get the sheet, but there is also an `assert!` to check if the sheet is properly deserialized. Therefore panics when it finds the sheet but is not deserialized.
I find this rather clunky, if functions has multiple failure modes, then it is optimal to make the return type as `Result` rather than an `Option`. Better we just let them handle failure cases than panicing user's code.
2. The code contains multiple instances of `String` type errors like "Not found.", "out of index.", "sheet not found.", "name duplicate.", "Index cannot be less than one.", "Non-standard range." etc. Rather than strings to signify the error, it would be wise to take advantage of types to classify and relay them to the user.(Also applicable for point-1 above)
3. I understand that the most of the `panic!` are associated with the reading of Zip file, but there is `ZipError` in the codebase which we can use.

Kindly comment your thoughts on my request for improving function signatures and removal of unnecessary panics.

MathNya:
@agentjill 
Thank you for your suggestion.
I agree with all 1-3.

However, for 1, the deserialization specification is unique, so I think it will be difficult to understand why it is None.
Therefore, I think it is necessary to add some additional information in the comments.

agentjill:
> However, for 1, the deserialization specification is unique, so I think it will be difficult to understand why it is None.
Therefore, I think it is necessary to add some additional information in the comments.

I understand the need to have normal and "lite" functionality (involving leaving out deserialisation). But my itention is not leave it for the user to read comments, rather have them handle the error possibility. Also, in my opinion this issue should not panic the user's code.

kindly share your thoughts on it. Also let me know if I have misunderstood your statement.

MathNya:
@agentjill 
Sorry, here I misunderstood.
Your suggestion to change the return value from Option to Result is perfectly fine.

c-git:
*TLDR* - What do we do about the panics? Change to results and change the API all over?

# Long version

[...]

PR #311, opened 2026-03-01
fix: handle backslash zip paths, missing styles.xml, and arithmetic overflows

developer0hye:
## Summary

This PR fixes **21 panicking XLSX files** from real-world sources (LibreOffice, Apache POI) by addressing 5 categories of issues. It contributes to the panic reduction effort proposed in #271 and resolves the specific files reported in #310.

**Note:** This branch is based on v2.3.3. The master branch has since been refactored to v3.0 with many structural changes, so this PR will have conflicts. The fixes are logically straightforward and can be cherry-picked or manually ported. I'm happy to rebase if needed.

### 1. Backslash zip path normalization (`reader/driver.rs`)

**Problem:** Some Windows tools generate XLSX files with backslash path separators (e.g., `xl\workbook.xml` instead of `xl/workbook.xml`) and lowercase `[content_types].xml`. `ZipArchive::by_name()` does exact string matching and fails.

**Fix:** Added `zip_by_name()` helper that tries:
1. Exact match (zero overhead for well-formed files)
2. Backslash-separated variant
3. Case-insensitive scan with separator normalization

All `arv.by_name()` calls replaced with `zip_by_name()`.

**Files fixed:** `tdf131575.xlsx`, `tdf76115.xlsx`, `49609.xlsx`

### 2. Non-standard XML namespace prefixes (`reader/driver.rs`)

**Problem:** Some XLSX files use non-standard namespace prefixes like `d3p1:id` instead of `r:id` for relationship attributes.

**Fix:** Extended `get_attribute()` to fall back to local-name matching when the exact prefixed key is not found.

### 3. Missing optional `styles.xml` (`reader/xlsx/styles.rs`)

**Problem:** `styles::read()` propagated `ZipError::FileNotFound` via `?`, but per the OOXML spec, `styles.xml` is optional.

**Fix:** Applied the same `FileNotFound` handling pattern used by `shared_strings.rs`, `doc_props_app.rs`, etc.

**Files fixed:** `56278.xlsx`, `tdf121887.xlsx`, `59021.xlsx`

### 4. Arithmetic overflow (`cell_formula.rs`, `coordinate.rs`, `number_formater.rs`)

**Problem:**
- `cell_formula.rs:210`: `self_col_num - root_col_num` underflows when `self < root`
- `coordinate.rs:120`: `num + offset_num` overflows after wrapping subtraction
- `number_formater.rs:137`: `10i32.pow(right.len())` overflows for large decimal strings

**Fix:**
- Use `wrapping_sub` for formula offset calculation
- Use `saturating_add`/`saturating_sub` in coordinate adjustment
- Widen `i32` to `i64` with `.min(18)` cap and `saturating_mul`
- Guard `index_from_coordinate` results with `let-else` patterns

**Files fixed:** `functions-excel-2010.xlsx`, `FormulaEvalTestData_Copy.xlsx`

### 5. `unwrap()` on None (`raw_relationship.rs`)

**Problem:** `get_attribute(e, b"Id").unwrap()` panics when relationship elements are missing required attributes.

**Fix:** Use `let-else` pattern to skip relationships with missing attributes.

**Files fixed:** `64450.xlsx`

## Test Results

- All 97 tests pass (88 existing + 9 new)
- `cargo fmt --all -- --check` passes
- `cargo clippy --workspace` passes
- 9 new test files added to `tests/test_files/`

Related: #310 #271


MathNya:
@developer0hye 
Thank you for the excellent PR.
I'll merge it into the master branch.
I'll handle the conflicting parts with a cherry-pick.

PR #315, opened 2026-04-27
Add jiff support

### LINKED ISSUE #313: PR Permission request for jiff support
Hi,

Thank you very much for this crate. I'm currently using the function excel_to_date_time_object and it turns a type from the chrono library. I'm currently transitioning to the jiff library and there are a few footguns doing the conversion so I'm going to take my time and see what's the best way. I wanted to find out if you'd be amenable to be submitting a PR with the final solution to help the next person. See link below to note from the maintainer of chrono about it's soft deprication.

https://github.com/chronotope/chrono/issues/1301#issuecomment-4321773137

### LINKED ISSUE #314: Doc tests disabled?
Hi,

I was going through the code and noticed that `doctest` are disabled. I was curious if there was a reason.

https://github.com/MathNya/umya-spreadsheet/blob/9b7e2bf7a2e21d1b3bb189191dfd5be169f2f31c/Cargo.toml#L43

### LINKED ISSUE #316: Make use of rstest
Hi,

I was looking for existing tests for dates and I noticed that there are some tests where the same thing is repeated many times. I think making use [rstest](https://docs.rs/rstest/latest/rstest/#creating-parametrized-tests) would be helpful. Not sure if it's worthwhile to update the existing tests like the one below

https://github.com/MathNya/umya-spreadsheet/blob/9b7e2bf7a2e21d1b3bb189191dfd5be169f2f31c/src/helper/number_format.rs#L247

but I'm thinking I'll make use of it for the date code I'm working on. Is it ok for me to add `rstest` as a dev dependency?

### LINKED ISSUE #317: Date formatting specifier availabe
The follow is based on the docs for [jiff](https://docs.rs/jiff/latest/jiff/fmt/strtime/index.html#conversion-specifications) and [chrono](https://docs.rs/chrono/latest/chrono/format/strftime/index.html#specifiers).

Unless I'm mistaken there are specifiers available for fractional seconds `%f`.

https://github.com/MathNya/umya-spreadsheet/blob/9b7e2bf7a2e21d1b3bb189191dfd5be169f2f31c/src/helper/number_format/date_formater.rs#L51-L52

c-git:
This PR includes:
- [x] Addresses or allows all clippy lints (361 lints, 357 the same)
- [x] Fixes style errors flagged by the beta compiler
- [x] Fix CI for `cargo fmt` to fail if differences are detected (was not actually failing on errors before)
- [x] Fix doctest so they compile properly
- [x] Corrections to doctest so they pass
- [x] Corrections to docs to match the code in date helpers module
- [x] Add support for jiff library
- [x] Correct code for existing conversion to chrono to match excel
- [x] Add tests to verify date conversions in both directions
- [x] Add support for formatting fractional seconds
- [ ] Deal with invalid code in examples

Please let me know if you'd like me to separate out any parts or remove the change all together or do it differently. I'm amenable to making corrections as desired.

Closes #313
Closes #314 
Closes #316 
Closes #317

Open question:
1. Should I put the two time libraries behind feature flags? (one for chrono and one for jiff).
2. Should we also add jiff and chrono to excel date/time?
3. Should we drop millisecond precision and round off to the nearest second? 
    - If we do round off to the nearest second instead then we have similar behaviour to the previous code. (went with this for now so the tests will pass)
    - If we do not we keep the full resolution but when we apply formatting it will show the actual seconds potion and the remainder (the fractions of a second) will not show if it's not part of the formatting but it seems when that's the case excel rounds but this will not and this will show different.

Example

```
failed: read_and_wite
 44180.584027777775, 12/15/2020 14:00
thread 'read_and_wite' (773707) panicked at tests/integration_test.rs:169:5:

assertion `left == right` failed
   left: "12/15/2020 14:01"
  right: "12/15/2020 14:00"
```

I used Excel from Office365 to generate the following list which I will be using to validate the code. I wasn't able to switch excel to use the 1904 Mac version so I just added 1462 days as can be seen in the formula bar.
image

PS. 
- I came across this [discussion](https://github.com/BurntSushi/jiff/discussions/383) while looking for how to convert between excel dates and jiff. It only covers jiff to excel and I don't know if that's something we want to provide support for.


c-git:
It's basically ready now. You'll let me know what you think please. Sorry it's so big. Let me know if you prefer I split it into separate PRs or if you want me to squash my changes into less commits

c-git:
Ok that's as good as I was able to do. I don't understand the remaining problems. There are calls to functions I just can't find in the code. I marked all the locations with TODO

MathNya:
@c-git 
Thank you.
I expect to have some free time around the day after tomorrow,
so I will proceed with the review.

MathNya:
@c-git 
Sorry for the delay; I’ve taken a look.

> Addresses or allows all clippy lints (361 lints, 357 the same)

Thank you for your assistance.
For some reason, it wasn’t detected on my end, and I was having trouble figuring out how to resolve it.
I apologize for the trouble this caused you.

> Should I put the two time libraries behind feature flags? (one for chrono and one for jiff).

No, I don’t think it’s necessary to go that far.
For the time being, I plan to keep both options available and eventually phase out Chrono.

> Should we also add jiff and chrono to excel date/time?

I'm sorry, but I couldn't quite grasp what this was about.
At the very least, since chrono will eventually be deprecated, I don't think support is necessary.

> Should we drop millisecond precision and round off to the nearest second?

I think it’s best to keep the milliseconds as they are without rounding them.
Even if this results in differences from Excel, it can’t be helped.
(There are other areas where differences from Excel arise, such as color settings. To be honest, I’ve given up on trying to make them match perfectly.)

c-git:
> Sorry for the delay; I’ve taken a look.

Let not your heart be troubled. You got to this quick. I have no claim on your time. I appreciate what you've done with this crate. I'm not currently blocked so it's not urgent for me. I am using the fork for the PR.

> Thank you for your assistance.
> For some reason, it wasn’t detected on my end, and I was having trouble figuring out how to resolve it.
> I apologize for the trouble this caused you.

Let not year heart be troubled maybe I'm using a newer version of rust but it was also failing in CI.



>> Should we also add jiff and chrono to excel date/time?

> I'm sorry, but I couldn't quite grasp what this was about.
> At the very least, since chrono will eventually be deprecated, I don't think support is necessary.

Yes I'm sorry I can see how that wasn't clear. I meant right now we provide a way to convert [from a excel date/time](https://github.com/MathNya/umya-spreadsheet/blob/313a3de54927bc7f0fa3116edde3d5e8de638976/src/helper/date.rs#L60) but to go in the other direction we require the parts like year month and so on. Would it make sense to add a `fn jiff_date_time_to_excel(value: jiff::civil::DateTime) -> f64` (see on of the existing function below, I'm proposing adding a new one).

https://github.com/MathNya/umya-spreadsheet/blob/313a3de54927bc7f0fa3116edde3d5e8de638976/src/helper/date.rs#L132-L139

> I think it’s best to keep the milliseconds as they are without rounding them.

Ok agreed, I will update the code to restore the precision and see how I can adjust the test to "work" and you'll see if that makes sense.

c-git:
Restored millisecond precision and updated the tests.

MathNya:
@c-git 
> Yes I'm sorry I can see how that wasn't clear. I meant right now we provide a way to convert [from a excel date/time](https://github.com/MathNya/umya-spreadsheet/blob/313a3de54927bc7f0fa3116edde3d5e8de638976/src/helper/date.rs#L60) but to go in the other direction we require the parts like year month and so on. Would it make sense to add a fn jiff_date_time_to_excel(value: jiff::civil::DateTime) -> f64 (see on of the existing function below, I'm proposing adding a new one).

Thank you.I understand.
Yes. I would definitely like to implement the `jiff_date_time_to_excel` function you suggested.
If possible, could you create this function for me?

c-git:
> @c-git
> 
> > Yes I'm sorry I can see how that wasn't clear. I meant right now we provide a way to convert [from a excel date/time](https://github.com/MathNya/umya-spreadsheet/blob/313a3de54927bc7f0fa3116edde3d5e8de638976/src/helper/date.rs#L60) but to go in the other direction we require the parts like year month and so on. Would it make sense to add a fn jiff_date_time_to_excel(value: jiff::civil::DateTime) -> f64 (see on of the existing function below, I'm proposing adding a new one).
> 
> Thank you.I understand. Yes. I would definitely like to implement the `jiff_date_time_to_excel` function you suggested. If possible, could you create this function for me?

100% just wanted to know if it was desirable. Will update PR.

c-git:
Implemented

c-git:
Might be a good idea for us to separate out the nightly clippy job as it's likely to fail somewhat often I think.

MathNya:
@c-git 
Thank you very much for your generous support.
I will merge these.

c-git:
My pleasure, thank you for the library

c-git:
Hi,

I just realized we never addressed the "broken" examples.

>  - [ ] Deal with invalid code in examples

For example this one 

https://github.com/MathNya/umya-spreadsheet/blob/40b34c965a2b1f864b61f86731cc014d3d2b5e11/src/lib.rs#L67

PR #320, opened 2026-05-08
Fix Deletion of Rows Containing Some Table Formula

WilliamTCarroll:
This is an update, as an unwrap() panic was tripped when using a table with formula referencing the same table (sometimes used for summing or calculating different values on the same table).

The Coordinate Regex was reading tokens like `Table1[[#This Row]`, as the Regex didn't have bounds for start and end.

By updating this regex to include a start (`^`) and end (`$`) of string bound:
```regex
^((\$)?([A-Z]{1,3}))?((\$)?([0-9]+))?$
```
It will still catch the cell references, but not catch any Table Reference as above.  All the Tests still pass.

There was also an issue with Ranges not properly reading when a Sheet Name was specified.

EX:
`Sheet 1!A1:B2`
Was split via `:`, leaving:
`Sheet1!A1` and `B2`
So, this was updated to split by the `!` first, and using the whole string if not found.

Thank you very much!


MathNya:
@WilliamTCarroll 
Thank you for the PR.
Since there were no issues with the content, I’ll go ahead and merge it.

PR #321, opened 2026-05-13
perf: memory perf, lazy read & stream read

attila-lin:
I encountered a fairly serious performance issue when processing large tables. In some cases, memory usage would continue growing and could even reach tens of gigabytes.

I used AI-assisted optimization to refactor the lazy-read path into a streaming-based approach. With this change, memory usage now stays at a much healthier level (around 50 MB in my test case).

That said, I have to admit I did not manually review every part of the generated code in depth.

If you notice any issues or have suggestions, please let me know and I’ll be happy to make further adjustments.

### DIFF
--- a/tests/integration_test.rs
+++ b/tests/integration_test.rs
@@ -4,13 +4,161 @@
 extern crate chrono;
 extern crate umya_spreadsheet;
 use std::{
-    io::Read,
+    env,
+    ffi::c_void,
+    fs::File,
+    io::{
+        BufWriter,
+        Read,
+    },
+    path::PathBuf,
     time::Instant,
 };
 
 use umya_spreadsheet::*;
 
 use crate::helper::color;
+
+#[cfg(target_os = "linux")]
+fn current_process_memory_mb() -> Option<f64> {
+    process_status_memory_mb("VmRSS:")
+}
+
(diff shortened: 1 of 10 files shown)


tests/integration_test.rs
```diff
@@ -115,6 +263,72 @@ fn lazy_read_and_wite_large_string() {
...
+            writer::xlsx::write_writer(&book, BufWriter::new(file)).unwrap();
+        }
+        Ok("light") => writer::xlsx::write_light(&book, &output_path).unwrap(),
+        _ => writer::xlsx::write(&book, &output_path).unwrap(),
+    }
+    print_large_file_checkpoint("write", total);
+}
+
+#[test]
+#[ignore]
```
c-git, line 309: Why are the test that were added ignored?

attila-lin: I use env to run the special perf. So we don't need to run it every time

c-git: I see, thanks for explaining.

MathNya:
@attila-lin 
Thank you for the PR.
I will review the details.

MathNya:
@attila-lin 
Since there were no issues, I'll merge it.

PR #323, opened 2026-05-19
feat: restore rounded second precision

### LINKED ISSUE #322: Time precision revert?
Hi,

I've been using the new version with the more precise time and I'm thinking we should revert back to how it was. For example I have an amount of hours that went from 12.0 to 11.999999166666667 and that's all over my app. It rounds off to the same thing but I think the lower precision is a better human experience. Is it ok with you if I switch it back? What do you think?

c-git:
I reverted the default function to give rounded seconds but added a millisecond version in case someone has that use case. However, I feel after trying it in practice for a while my users did not appreciate the added precision. In case that's different for someone else, I added an alternate version.

Closes #322

### DIFF
--- a/src/helper/date.rs
+++ b/src/helper/date.rs
@@ -73,7 +73,84 @@ pub fn excel_to_date_time_chrono(excel_timestamp: f64) -> NaiveDateTime {
         + Duration::seconds(cast(seconds).unwrap())
 }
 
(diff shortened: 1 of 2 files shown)


MathNya:
@c-git 
Thank you. I'll merge it.

PR #326, opened 2026-06-04
fix(alignment): persist indent attribute on read/write

### LINKED ISSUE #327: alignment.indent attribute dropped on read→write round-trip
## Bug

The `indent` attribute on `<alignment>` elements in `xl/styles.xml` is silently dropped on any read→write round-trip through umya. The `Alignment` struct has no field for it, so:

- The reader ignores the attribute on parse.
- The writer never emits one.
- Cells authored via the umya API have no way to set an indent.
- Cells read from an external xlsx (Excel, LibreOffice, openpyxl) that have a non-zero indent lose it on save.

The `indent` attribute is part of the OOXML spec ([ECMA-376 §18.8.1](https://learn.microsoft.com/en-us/openspecs/office_standards/ms-xlsx/), `CT_CellAlignment`) and Excel itself round-trips it, so this is observable as data loss when umya is used in a pipeline that reads an authored xlsx and saves it back.

## Reproduction

1. Create a small xlsx via openpyxl (or Excel) with one cell whose alignment has `indent=3`:
   ```python
   from openpyxl import Workbook
   from openpyxl.styles import Alignment
   wb = Workbook()
   wb.active['A1'].alignment = Alignment(horizontal='left', indent=3)
   wb.save('input.xlsx')
   ```
   The resulting `xl/styles.xml` contains `<alignment horizontal="left" indent="3"/>`.

2. Round-trip it through umya:
   ```rust
   let book = umya_spreadsheet::reader::xlsx::read("input.xlsx")?;
   umya_spreadsheet::writer::xlsx::write(&book, "output.xlsx")?;
   ```

3. Inspect the output's `xl/styles.xml`:
   ```
   <alignment horizontal="left"/>
   ```
   The `indent="3"` is gone.

## Discovery

Found via a Rust workbook builder that uses umya in [`formualizer`](https://github.com/formualizer/formualizer)'s recalc path. The build authored cells with indent attributes via `rust_xlsxwriter`, and the recalc round-trip stripped every one of them — observable as broken indentation in the output xlsx.

## Fix

PR #326 — adds the `indent` field plus its API surface, hash contribution, reader entry, and writer emission, following the conventions of the sibling `Alignment` fields (`horizontal`, `vertical`, `wrap_text`, `text_rotation`). Includes a runtime regression test using a committed fixture xlsx authored by openpyxl, confirmed to panic on the unfixed code with `indent attribute missing from round-tripped styles.xml`.

jon-convexent:
Fixes #327.

## Summary

The `Alignment` struct lacks an `indent` field, so the `indent` attribute on `<alignment>` elements in `xl/styles.xml` is silently dropped on any read→write round-trip. This means cells authored via the umya API with a non-zero indent have no way to express it, and cells read from upstream xlsx files with an indent set lose it on save.

This PR adds first-class support for `alignment.indent`, mirroring the conventions established by the sibling fields (`horizontal`, `vertical`, `wrap_text`, `text_rotation`):

- New field `indent: UInt32Value` on `Alignment`
- Getter `indent()` (with `#[must_use]`) + deprecated `get_indent()` shim
- Setter `set_indent(u32)`
- Hash contribution via `indent.hash_string()`
- Read-side `set_string_from_xml!(self, e, indent, "indent")`
- Write-side attribute emission when `has_value()`

## Discovery

Surfaced from a downstream Rust workbook builder that uses umya through the recalc path of [`formualizer`](https://github.com/formualizer/formualizer). The build authored cells with indents 1–3 via `rust_xlsxwriter`, then `formualizer` opened the file with umya, recalculated formulas, and wrote it back via `umya_spreadsheet::writer::xlsx::write`. The pre-recalc xlsx had four `indent="N"` attributes in `styles.xml`; the post-recalc xlsx had zero. Tracing it to umya's `Alignment::write_to` showed the `indent` attribute was never emitted because the field didn't exist.

## Test plan

Two tests in `tests/integration_test.rs`:

### `alignment_indent_preserved_from_external_xlsx` — bug reproduction

Reads a small fixture xlsx (`tests/test_files/alignment_indent_input.xlsx`, authored by openpyxl) whose `xl/styles.xml` contains `<alignment horizontal="left" indent="3"/>`, writes it back via umya, then inspects the round-tripped `xl/styles.xml` and asserts it still contains `indent="3"`.

Confirmed to panic on the unfixed code:

```
thread 'alignment_indent_preserved_from_external_xlsx' panicked at tests/integration_test.rs:2727:5:
indent attribute missing from round-tripped styles.xml; got:
...<alignment horizontal="left"/>...
```

After the fix the output contains `<alignment horizontal="left" indent="3"/>` and the test passes.

### `alignment_indent_roundtrip` — API surface

Sets `indent=3` via the new `set_indent` API, writes, re-reads, and asserts `alignment.indent() == 3`. Confirmed to compile-fail on the unfixed code (no `set_indent` / `indent()` methods on `Alignment`), so it also guards against a regression that re-added the indent field but forgot the API surface.

```
running 2 tests
test alignment_indent_preserved_from_external_xlsx ... ok
test alignment_indent_roundtrip ... ok

test result: ok. 2 passed; 0 failed
```

## Notes

- Following the project's documented style: new accessor names without the `get_` prefix, deprecated `get_*` shim with `since = "3.0.0"`, `#[must_use]` on the value-returning getter.
- No public API removed; only additions.
- No xlsx schema changes — the `indent` attribute on `<alignment>` is part of the OOXML spec ([ECMA-376 §18.8.1](https://learn.microsoft.com/en-us/openspecs/office_standards/ms-xlsx/), `CT_CellAlignment`) and Excel itself round-trips it.


MathNya:
@jon-convexent 
Thank you for the PR.
Since there were no issues, I'll merge it.

PR #329, opened 2026-06-08
Fix #328, panic when writing a workbook to a file without an extension.

### LINKED ISSUE #328: Panic when writing a workbook to a file without an extension.
Hi, I encountered a panic when trying to write a workbook to a temporary file using this library. After investigation, it turned out to be caused by the temporary file lacking a file extension. I think this behavior is not correct. I'd like to confirm whether the library, as designed, assumes that files without an extension should not be written to.

# Symptom
Panic when writing a workbook to a file without an extension.
```
   Compiling minimum_test v0.1.0 (E:\TmpProjects\minimum_test)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.87s                                                                                   
     Running `target\debug\minimum_test.exe`

thread 'main' (19016) panicked at E:\TmpProjects\umya-spreadsheet\src\writer\xlsx.rs:239:47:
called `Option::unwrap()` on a `None` value
stack backtrace:
   0: std::panicking::panic_handler
             at /rustc/ac68faa20c58cbccd01ee7208bf3b6e93a7d7f96/library\std\src\panicking.rs:689
   1: core::panicking::panic_fmt
             at /rustc/ac68faa20c58cbccd01ee7208bf3b6e93a7d7f96/library\core\src\panicking.rs:80
   2: core::panicking::panic
             at /rustc/ac68faa20c58cbccd01ee7208bf3b6e93a7d7f96/library\core\src\panicking.rs:150
   3: core::option::unwrap_failed
             at /rustc/ac68faa20c58cbccd01ee7208bf3b6e93a7d7f96/library\core\src\option.rs:2251
   4: enum2$<core::option::Option<ref$<std::ffi::os_str::OsStr> > >::unwrap
             at C:\Users\Alex\.rustup\toolchains\stable-x86_64-pc-windows-msvc\lib\rustlib\src\rust\library\core\src\option.rs:1016
   5: umya_spreadsheet::writer::xlsx::write<ref$<std::path::Path> >
             at E:\TmpProjects\umya-spreadsheet\src\writer\xlsx.rs:239
   6: minimum_test::main
             at .\src\main.rs:6
   7: core::ops::function::FnOnce::call_once<void (*)(),tuple$<> >
             at C:\Users\Alex\.rustup\toolchains\stable-x86_64-pc-windows-msvc\lib\rustlib\src\rust\library\core\src\ops\function.rs:250
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
error: process didn't exit successfully: `target\debug\minimum_test.exe` (exit code: 101)
```

# Minimum Code
```rust
use umya_spreadsheet;

fn main() {
    let book = umya_spreadsheet::new_file();
    let path = std::path::Path::new("test_without_extension");
    let _unused = umya_spreadsheet::writer::xlsx::write(&book, path);
}
```

# Reason
At [line 239](https://github.com/MathNya/umya-spreadsheet/blob/a8e9ccbe2e2ba324522afd95ec31185381996eb6/src/writer/xlsx.rs#L239) of the file located at writer/xlsx.rs, the code uses `path.as_ref().extension().unwrap().to_str().unwrap()` to retrieve the file extension, assuming that the file always has an extension. This leads to a panic due to the use of unwrap.

# Suggestion
I think there are several different ways to fix it, depending on the specific situation.
- Use `unwrap_or` to create tmp file with default extension;
- Just return Err if the file has no extension.
- Just append the .tmp suffix directly to the path.

AlexZhu2001:
See issue #328 for detail.
Add default extension to temporary file if output file with no extension.

### DIFF
--- a/src/writer/xlsx.rs
+++ b/src/writer/xlsx.rs
@@ -236,7 +236,11 @@ pub fn write_writer_light<W: io::Write>(wb: &Workbook, mut writer: W) -> Result<
 /// let _unused = umya_spreadsheet::writer::xlsx::write(&book, path);
 /// ```
 pub fn write<P: AsRef<Path>>(wb: &Workbook, path: P) -> Result<(), XlsxError> {
-    let extension = path.as_ref().extension().unwrap_or("xlsx").to_str().unwrap_or("xlsx");
+    // Fix #328, add default extension
+    let extension = match path.as_ref() {
+        Some(ext) => ext.to_str().unwrap_or("xlsx"),
+        None => "xlsx",
+    };
     let path_tmp = path
         .as_ref()
         .with_extension(format!("{}{}", extension, "tmp"));
--- a/src/writer/xlsx.rs
+++ b/src/writer/xlsx.rs
@@ -237,7 +237,7 @@ pub fn write_writer_light<W: io::Write>(wb: &Workbook, mut writer: W) -> Result<
 /// ```
 pub fn write<P: AsRef<Path>>(wb: &Workbook, path: P) -> Result<(), XlsxError> {
     // Fix #328, add default extension
-    let extension = match path.as_ref() {
+    let extension = match path.as_ref().extension() {
         Some(ext) => ext.to_str().unwrap_or("xlsx"),
         None => "xlsx",
     };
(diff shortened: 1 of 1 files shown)


MathNya:
@AlexZhu2001 
Thank you for the shout-out.
That’s a great idea.

AlexZhu2001:
My pleasure.😊

PR #331, opened 2026-06-15
fix(formula): translate-to-origin semantics for shared-formula siblings

jon-convexent:
## Problem

Shared-formula sibling cells — `<f t="shared" si="N"/>` with no inline text, sharing a master via `si` — are expanded incorrectly. The master's formula should be re-based onto each sibling by translating every **relative** reference by `(sibling_col - master_col, sibling_row - master_row)` (absolute refs like `$A$1` stay put). Instead, every sibling ends up referencing the **master's** relative cells.

Concretely, for `A1=10`, master `B1==A1+1` shared across `B1:F1`:

| cell | expected | actual (before) |
|------|----------|-----------------|
| C1   | `B1+1`   | `A1+1`          |
| D1   | `C1+1`   | `A1+1`          |
| …    | …        | `A1+1`          |

## Root cause

`CellFormula::set_attributes` expands shared siblings with `adjustment_insert_formula_coordinate`, which has **insert-shift** semantics ("shift references at >= root by N", used when a row/column is inserted). That is the wrong primitive for re-basing relative refs to a new origin — it leaves refs below/left of the root untouched, so siblings collapse to the master's references.

## Fix

Use `adjustment_formula_coordinate` + `render` — the translate-to-new-origin primitive the crate already uses in `Cell::set_coordinate`. Signed offsets are built via `num_traits::cast` to match that existing caller.

## Test

Adds `shared_formula_siblings_rebase_relative_references`: builds a 5-cell shared chain in memory, round-trips through write+read, and asserts each sibling re-based. It **fails on the current code** (`C1` reads `A1+1`) and passes with this change.

## Checks

`cargo +nightly fmt`, `cargo clippy --all-targets -- -D warnings`, and `cargo test --test integration_test` (full suite) all pass.

### DIFF
--- a/src/structs/cell_formula.rs
+++ b/src/structs/cell_formula.rs
@@ -17,9 +17,11 @@ use crate::{
         coordinate::index_from_coordinate,
         formula::{
             FormulaToken,
+            adjustment_formula_coordinate,
             adjustment_insert_formula_coordinate,
             adjustment_remove_formula_coordinate,
             parse_to_tokens,
+            render,
         },
     },
     reader::driver::{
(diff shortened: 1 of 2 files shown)


MathNya:
@jon-convexent 
Thank you for the PR.

PR #332, opened 2026-06-15
fix(chart): preserve absence of <c:legend> on read/write round-trip

jon-convexent:
## Problem

A chart whose source xlsx has **no** `<c:legend>` element gains one on a read→write round-trip. `Chart::write_to` calls `self.legend.write_to(writer)` unconditionally, and `Legend::write_to` always emits the `<c:legend>` block — there is no notion of "this chart has no legend." This silently undoes, e.g., `rust_xlsxwriter`'s `chart.legend().set_hidden()` on any file that passes through umya.

## Fix

Track legend presence on `Chart`:
- `legend_present: bool`, set to `true` on read only when a `<c:legend>` element is actually seen (reset to `false` at the start of `set_attributes`).
- `write_to` emits the legend only when `legend_present`.
- Defaults to `true` for programmatically-constructed charts (`Chart::default()`), and `legend_mut()` / `set_legend()` flip it to `true`, so existing code that builds or mutates a legend is unaffected.
- Adds `legend_present()` / `set_legend_present()` — the latter is the public "no legend" toggle (matches Excel / `rust_xlsxwriter` `set_hidden()`).

## Tests

- `chart_legend_absence_roundtrips` — builds a no-legend fixture (strips `<c:legend>` from `aaa.xlsx`'s chart parts), reads + writes it, and asserts no legend reappears. **Fails on the current code** (11/11 charts regain a legend).
- `set_legend_present_false_suppresses_legend` — exercises the new public API.
- Existing charts-with-legends continue to round-trip with their legends (full suite green).

## Checks

`cargo +nightly fmt`, `cargo clippy --all-targets -- -D warnings`, and `cargo test --test integration_test` (full suite) all pass.

### DIFF
--- a/src/structs/drawing/charts/chart.rs
+++ b/src/structs/drawing/charts/chart.rs
@@ -33,7 +33,7 @@ use crate::{
     xml_read_loop,
 };
 
-#[derive(Clone, Default, Debug)]
+#[derive(Clone, Debug)]
 pub struct Chart {
     title:                         Option<Title>,
     auto_title_deleted:            AutoTitleDeleted,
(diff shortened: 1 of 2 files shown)


MathNya:
@jon-convexent 
Thank you for this PR as well.
I’ll merge it.

PR #342, opened 2026-07-18
fix: preserve numeric-looking inline strings

developer0hye:
## Summary

- Preserve `inlineStr` cell values as strings instead of guessing their type.
- Return string and rich-text cells unchanged from `formatted_value()` instead of applying numeric number-format coercion.
- Add parser-level regression coverage for a numeric-looking inline string (`0050`) through both raw and formatted value APIs.

## Root cause

The XLSX reader routed `inlineStr` text through `set_value_crate`, which guesses scalar types. That converted values such as `0050` to the number `50`, losing both the leading zeros and the source cell type. After preserving the raw string type, `formatted_value()` still passed the text through the number formatter, which again normalized it to `50` under the General format.

The reader now uses the existing `set_value_string_crate` path, and formatted text values bypass numeric formatting just as Excel text cells do.

## Impact

Numeric-looking text stored as OOXML inline or shared strings now remains text through both value APIs. Numeric cells continue to use their number formats.

## Validation

- `cargo test` (86 library tests, 112 integration tests, 4 streaming-writer tests, and 79 doc tests passed; 23 tests ignored)
- `cargo clippy --lib -- -D warnings`

### DIFF
--- a/src/structs/cell.rs
+++ b/src/structs/cell.rs
@@ -651,7 +651,7 @@ impl Cell {
                     },
                     b"is" => {
                         if type_value == "inlineStr" {
-                            self.set_value_crate(&string_value);
+                            self.set_value_string_crate(&string_value);
                         }
                     }
                     b"c" => return,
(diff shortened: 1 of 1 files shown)


c-git:
I legit had this problem this morning. Thanks. I had some text with leading zeros and my zeros got removed.

developer0hye:
@c-git cool

MathNya:
@developer0hye 
Thank you for the PR.

PR #345, opened 2026-07-21
fix(number_format): round decimal display values half away from zero like Excel

developer0hye:
## Problem

`get_formatted_value()` disagrees with Excel's rendering for three common cases:

1. **Integer formats truncate instead of rounding.** `107310.6` with `#,##0` returns `107,310`; Excel shows `107,311`. `format_straight_numeric_value` returned the integer block verbatim whenever the format had no decimals.
2. **Short fractions are multiplied instead of zero-padded.** `39.1` with `0.00` returns `39.100`: the right-side digits were parsed as an integer and multiplied by `10^width` (`1 × 100 = 100`) rather than padded to `10`.
3. **Percentage ties round the wrong way.** `1.065` with `0%` returns `106%`; Excel shows `107%`. `1.065 × 100` in binary doubles is `106.4999…`, so anything that multiplies first loses the tie. Excel rounds the decimal display value.

## Fix

Round the decimal string directly instead of routing values back through binary floats:

- New `round_decimal_string` rounds a plain decimal string (as produced by `f64::to_string`, i.e. shortest round-trip) half away from zero at the format's fractional width, zero-padding short fractions, with digit-carry propagation for `999.95`-style rollovers. Exponent-notation input falls back to float formatting.
- New `group_thousands` inserts comma separators into the integer part by string, so grouping survives rounding carries (`999,999.5` with `#,##0` → `1,000,000`) and values beyond integer range keep their digits.
- `format_as_percentage` now multiplies by shifting the decimal point in the string (`1.065` → `106.5`) before rounding, which preserves decimal ties exactly.

## Testing

- New unit tests in `number_formater.rs` covering half-away rounding (positive/negative), zero-padding, excess-decimal rounding, currency prefixes, and thousands grouping through rounding carries, plus tie-rounding tests in `percentage_formater.rs`
- `cargo test` — full suite green (94 + 112 + 79 + 4 tests)

### DIFF
--- a/src/helper/number_format/number_formater.rs
+++ b/src/helper/number_format/number_formater.rs
@@ -1,7 +1,5 @@
 use std::borrow::Cow;
 
-use thousands::Separable;
-
 use super::fraction_formater::format_as_fraction;
 use crate::helper::utils::compile_regex;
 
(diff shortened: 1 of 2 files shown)


c-git:
@developer0hye would you be able to address the two failing tests? The fix doesn't look like it should be too hard. One looks like just to run cargo fmt with nightly rust and the other is just to fix a format string. If you're getting trouble let me and I can give it a try, shouldn't be too hard.

developer0hye:
@c-git Thanks for the pointers — both were exactly as you described. Fixed in 865316ab:

- **formatting**: ran nightly `cargo fmt` (reformatted the `if int_digits.is_empty()` block in `number_formater.rs` and the `use` ordering in `percentage_formater.rs`).
- **build/clippy**: the `uninlined-format-args` lint — changed `format!("{:.*}", decimals, parsed)` to `format!("{parsed:.decimals$}")`.

All checks are green now (build stable/beta/nightly/1.88.0 + formatting). No changes to the actual rounding logic. Thanks for the review!

c-git:
I didn't get a chance to actually understand the code. I just noticed that those were easy fixes.

MathNya:
@developer0hye 
Thank you.
Since there were no issues with this either, I'll merge it.

PR #347, opened 2026-07-21
fix: read dataBar cfvo and color children written as start/end tags

developer0hye:
## Summary

`DataBar::set_attributes` only matches `Event::Empty` for its `<cfvo>` and `<color>` children. Files written by producers that emit these as start/end pairs (`<color rgb="FF1E2761"></color>` — openpyxl among others) load with empty `cfvo_collection` and `color_collection`, so the data bar's color is silently lost.

`ColorScale::set_attributes` already handles both event kinds; this PR applies the same pattern to `DataBar`.

## Changes

- Match `Event::Start` in addition to `Event::Empty` for `cfvo`/`color` children of `<dataBar>`, passing the empty flag through to the child readers.
- Unit tests covering both the start/end-tag form and the self-closing form.

## Testing

- `cargo test` — full suite green (87 + 112 + 4 + 79 passed).
- New tests: `structs::data_bar::tests::read_child_elements_with_end_tags`, `read_child_elements_self_closing`.

### DIFF
--- a/src/structs/data_bar.rs
+++ b/src/structs/data_bar.rs
@@ -85,16 +85,17 @@ impl DataBar {
     ) {
         xml_read_loop!(
             reader,
-            Event::Empty(ref e) => {
+            ref n @ (Event::Empty(ref e) | Event::Start(ref e)) => {
+                let is_empty = matches!(n, Event::Empty(_));
                 match e.name().into_inner() {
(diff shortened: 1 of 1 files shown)


MathNya:
@developer0hye 
Thank you.
Since there were no issues with this either, I'll merge it.

Issue #336, opened 2026-07-01
Cell value return changed


c-git:
Hi,

Sorry to only report a problem without a cause or solution but don't have the bandwidth to dig into it right now but wanted to at least make you aware of it.

TLDR between [this commit](https://github.com/MathNya/umya-spreadsheet/commit/32337ffd70119b10b32d4bf59d19da11e937da80) and the current [head of master](https://github.com/MathNya/umya-spreadsheet/tree/4f5dc6a1f67b74418c87aa5399fef1626f9009ee) something caused the returned value of the cell to no longer be the same. I knocked together a small test to show the change but haven't had time to investigate more than what I've already shared.

Previously returned value (which matches what I see when I open the file in LibreOffice): `"{\"year_month\":{\"year\":2024,\"month\":11},\"period_number\":1}"`
Currently returned value: `":1}"`

Sample file used in the test [temp.xlsx](https://github.com/user-attachments/files/29563253/temp.xlsx)

Sample test code:

```rust
#[test]
fn debug_cell_value_change() {
    let path = std::path::Path::new("./tests/test_files/temp.xlsx");
    let mut book = umya_spreadsheet::reader::xlsx::read(path).unwrap();
    let sheet = book.sheet_by_name("Sheet1").unwrap();
    let value = sheet.cell((1, 2)).unwrap().value();
    assert_eq!(
        value,
        "{\"year_month\":{\"year\":2024,\"month\":11},\"period_number\":1}"
    );
}
```

Test output:
```
 1  failed: debug_cell_value_change
 thread 'debug_cell_value_change' (213485) panicked at tests/integration_test.rs:2946:5:
 assertion `left == right` failed
   left: ":1}"
  right: "{\"year_month\":{\"year\":2024,\"month\":11},\"period_number\":1}"
```

Ephemeral branch with sample code and file https://github.com/c-git/umya-spreadsheet/tree/debug_after (Not planning to keep this branch long term but it only has the code and file above)

MathNya:
@c-git 
Thank you for letting us know.
I’ll look into it as well.

c-git:
Confirmed fixed

PR #348, opened 2026-07-22
Fix formula text truncated on XML character references

sentinelt:
## Summary

`dataValidation` / formula character data that contains XML numeric character references (e.g. `&#8211;` for an en-dash) was truncated when reading `.xlsx` files.

### Root cause

quick-xml emits those references as `Event::GeneralRef` nodes between successive `Event::Text` fragments. The parsers for `DataValidation` (`formula1` / `formula2`) and `Formula` **replaced** the accumulator on each `Text` event and ignored `GeneralRef`, so only the last text fragment survived.

Example list formula in a real workbook:

```xml
<formula1>"0&#8211;3 years ago,4&#8211;5 years ago,6&#8211;7 years ago"</formula1>
```

was read as:

```text
7 years ago"
```

instead of the full list string.

### Fix

- Append successive `Text` events instead of overwriting
- Resolve and append `GeneralRef` (numeric char refs + common named entities)
- Shared helpers: `append_xml_text` / `append_xml_general_ref`
- Unit tests covering en-dash list formulas and plain list formulas
- Close a missing brace in `number_formater` tests that broke compilation of the local tree

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# Unreleased
+### Bug Fixed
+* Data validation / formula character data containing XML numeric character references (e.g. `&#8211;`) was truncated. `formula1`/`formula2` text is now accumulated across successive `Text` and `GeneralRef` events from quick-xml.
+
 # Change Detail -> 3.0.0
 ### We have changed the name of the Getter.
 ```rust
--- a/src/helper/number_format/number_formater.rs
+++ b/src/helper/number_format/number_formater.rs
@@ -344,6 +344,7 @@ mod tests {
     #[test]
     fn format_as_number_wraps_parenthesized_sections() {
         assert_eq!(format_as_number(1234.0, "(#,##0)"), "(1,234)");
+    }
 
     #[test]
     fn format_as_number_rounds_half_away_from_zero() {
--- a/src/helper/utils.rs
+++ b/src/helper/utils.rs
@@ -21,6 +21,42 @@ pub(crate) fn unescape_xml_text(e: &quick_xml::events::BytesText<'_>) -> String 
         .into_owned()
 }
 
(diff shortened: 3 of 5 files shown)


MathNya:
@sentinelt 
Thank you for the PR.
(I’ll make a few minor adjustments after merging.)

There are a few other instances of this, so it looks like we’ll need to review the entire code.

c-git:
@MathNya after you've done the update in the other places as well, please let me know so I can test if it also fixes my issue in https://github.com/MathNya/umya-spreadsheet/issues/336

MathNya:
@c-git 
We have addressed all the issues.
Please check #336.

c-git:
Thank you. I can confirm this fixed my problem. Sorry I never got time to work on it. Been under pressure.

PS. The hash of the output xlsx file has changed but I wasn't able to easily find the difference. It still has the some contents when saved as csv. I recall there was a tool you used to compare the files but I can't recall what it was and can't remember in which pr or issue you'd spoken about it. Would you mind sharing the tool again so I can look into what changed in the file please.

c-git:
FYI I extracted the xml from the zip, pretty printed it, and diffed it in vscode and found the change. It was to a column width. Would still love to know what tool you use cuz I'm sure that would have been faster but just FYI since I mentioned that something changed just reporting back.

image

MathNya:
@c-git 
Thank you for your prompt response.

I’m currently using this VSCode extension to check XML.
https://marketplace.visualstudio.com/items?itemName=yuenm18.ooxml-viewer

(There used to be a very user-friendly Google Chrome extension, but it’s now deprecated and can no longer be used.)

c-git:
> I’m currently using this VSCode extension to check XML. https://marketplace.visualstudio.com/items?itemName=yuenm18.ooxml-viewer

Thank you very much

PR #350, opened 2026-07-24
Fix shared formula expansion for whole-row and whole-column references

### LINKED ISSUE #349: Panic when expanding shared formulas with whole-column references
## Description

Reading an XLSX/XLSM workbook that contains a shared formula with a whole-column reference such as `A:B` panics while the reader expands the shared formula.

This reproduces with `umya-spreadsheet` 3.0.1 and current `master` (`566f99f62731bc79a568a8184410197855d0d75e`).

## Minimal workbook structure

The relevant worksheet XML is equivalent to:

```xml
<row r="3">
  <c r="H3">
    <f t="shared" ref="H3:H10" si="0">IF(G3="","",VLOOKUP(G3,A:B,2,FALSE))</f>
  </c>
</row>
<row r="4">
  <c r="H4"><f t="shared" si="0"/></c>
</row>
```

Reading the workbook is enough to trigger the panic:

```rust
let path = std::path::Path::new("shared-formula.xlsx");
let _book = umya_spreadsheet::reader::xlsx::read(path).unwrap();
```

## Actual result

```text
called `Option::unwrap()` on a `None` value
src/helper/formula.rs:893:46
```

The panic occurs in `adjustment_formula_coordinate`. For a whole-column endpoint such as `A`, `index_from_coordinate("A")` returns a column but no row. The code checks only that the column exists and then unconditionally unwraps the row.

Whole-row references such as `1:3` have the symmetric behavior: they do not enter this branch and therefore are not adjusted vertically.

## Expected result

Shared-formula expansion should:

- adjust normal cell references in both applicable axes;
- adjust whole-column references only in the column axis;
- adjust whole-row references only in the row axis;
- respect absolute (`$`) locks; and
- leave named ranges or other non-coordinate operands unchanged.

This is related to #216, but that fix only guards a missing column and does not handle column-only references with a missing row.

A tested fix with regression coverage is available at JunWang-Lin/umya-spreadsheet@be03d1b2c31e238632a2822a6af041fcb89588a4.

JunWang-Lin:
Closes #349.

## Summary

- avoid panicking while expanding shared formulas that contain whole-column or whole-row references
- adjust whole-column references only along the column axis and whole-row references only along the row axis
- preserve absolute row/column locks and named ranges
- produce `#REF!` when an adjusted reference moves out of bounds
- add regression tests for cell, whole-column, and whole-row references

## Root cause

Shared-formula expansion assumed every parsed reference had both a row and a column coordinate. Valid formulas such as `VLOOKUP(G3,A:B,2,FALSE)` contain a whole-column reference with no row coordinate, so the expansion path unwrapped `None` and panicked.

## Validation

- `cargo +nightly fmt`
- `cargo clippy --lib -- -D warnings`
- `cargo test` (299 passed, 23 ignored)
- verified the original workbook formula `IF(G3="","",VLOOKUP(G3,A:B,2,FALSE))`

## Note

The second commit restores a missing `}` in the existing `number_formater` test module on the current upstream `master`; without it, the repository cannot be formatted or compiled for validation.

### DIFF
--- a/src/helper/formula.rs
+++ b/src/helper/formula.rs
@@ -9,6 +9,7 @@ use crate::{
             adjustment_remove_coordinate,
             coordinate_from_index_with_lock,
             index_from_coordinate,
+            string_from_column_index,
         },
         range::{
             get_join_range,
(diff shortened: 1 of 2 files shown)


MathNya:
@JunWang-Lin 
Thank you for the PR.

PR #351, opened 2026-07-25
removed unnecessary collect to improve the memory utilisation.

agentjill:
hey @MathNya, in this PR i have tried to reduce the memory usage my reducing the collection of iterator into `Vec` as much as possible and lazily evaluate as many values required. This may slightly reduce the memory that the program uses. Also I have added some changes to internal function signature to be more generic.

### DIFF
--- a/src/helper/number_format.rs
+++ b/src/helper/number_format.rs
@@ -161,13 +161,10 @@ fn split_format(sections: Vec<&str>, value: f64) -> (String, String, String) {
 
         // Process color matching
         if let Some(captures) = color_re.captures(section).ok().flatten() {
-            let items: Vec<&str> = captures
-                .iter()
-                .filter_map(|cap| cap.map(|c| c.as_str()))
-                .collect();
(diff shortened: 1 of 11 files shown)


MathNya:
@agentjill 
Thank you for the PR.

agentjill:
@MathNya, the tests fail with following error

>error: this argument is passed by value, but not consumed in the function body
   --> src/structs/chart.rs:335:33
    |
335 |         area_chart_series_list: Vec<&str>,
    |                                 ^^^^^^^^^ help: consider changing the type to: `&[&str]`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/rust-1.97.0/index.html#needless_pass_by_value
    = note: `-D clippy::needless-pass-by-value` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(clippy::needless_pass_by_value)]`
    
 Here the function in reference `new_chart(..)` is a `pub fn`, if changes as asked by clippy are implemented then it would break the API surface. Kindly resolve this either denying the lint for this function or accepting the breaking change.

PR #357, opened 2026-08-24
Preserve t="str" on empty-string formula results

### LINKED ISSUE #356: [BUG] empty string cells are rendered as empty without a type
# Empty-string formula results lose their `t="str"` type on read/write round-trip

## Summary

When umya reads a formula cell whose cached result is an **empty string** — written by Excel/LibreOffice as a self-closing `<v/>` with `t="str"` — the string type is discarded. The cell's data type becomes empty (`""`) instead of `"s"`, and on write the `t="str"` attribute is dropped entirely.

Downstream this silently turns an empty-string formula result (`=IF(cond,"","x")` evaluating to `""`) into an untyped/blank cell. Tools that recalculate or read the file then treat the result as blank/number instead of an empty string, which changes calculation and text-vs-null semantics.

Non-empty string results (e.g. `<v>x</v>`) are **not** affected — only empty ones.

## Environment

- `umya-spreadsheet` 3.1.0 (latest release)
- `rustc` 1.95.0

## Minimal reproduction

`input.xlsx` (attached) contains a single formula cell whose cached result is an empty string, exactly as Excel/LibreOffice emit it:

```xml
<c r="A1" t="str"><f>IF(TRUE,"","x")</f><v/></c>
```

[input.xlsx](https://github.com/user-attachments/files/31371589/input.xlsx)

`Cargo.toml`:

```toml
[dependencies]
umya-spreadsheet = "3.1.0"
```

`src/main.rs`:

```rust
use std::path::Path;

fn main() {
    let book = umya_spreadsheet::reader::xlsx::read(Path::new("input.xlsx")).unwrap();
    let cell = book.sheet(0).unwrap().cell((1, 1)).unwrap();
    // input.xlsx A1: <c r="A1" t="str"><f>IF(TRUE,"","x")</f><v/></c>
    println!("formula   : {:?}", cell.formula());
    println!("data_type : {:?}", cell.data_type()); // "s" expected
    println!("value     : {:?}", cell.value());
    umya_spreadsheet::writer::xlsx::write(&book, Path::new("out.xlsx")).unwrap();
}
```

## Actual behavior

```
formula   : "IF(TRUE,\"\",\"x\")"
data_type : ""          <- should be "s", actual is ""
value     : ""
```

`out.xlsx` A1 (type attribute gone):

```xml
<c r="A1"><f>IF(TRUE,"","x")</f><v/></c>
```

## Expected behavior

The cached-result type is preserved across the round-trip:

- `data_type()` returns `"s"`
- `out.xlsx` A1 keeps `t="str"`: `<c r="A1" t="str"><f>IF(TRUE,"","x")</f><v/></c>`

## Notes on the cause

Two related details:

1. `CellValue::guess_typed_data("")` maps the empty string to `CellRawValue::Empty`, so a cached string result of `""` loses the fact that it was a string (`data_type()` -> `""` rather than `"s"`).
2. The distinction is tied to how the `<v>` element arrives from the parser. A self-closing `<v/>` (an `Event::Empty`, which is what Excel/LibreOffice write for empty values) drops the type, whereas an explicit `<v></v>` round-trips correctly. So a file authored by umya itself (which writes `<v></v>`) does not reproduce this; a file from Excel/LibreOffice does.

## Impact

Any pipeline that round-trips real-world spreadsheets through umya (read -> modify -> write, then recalculate/parse elsewhere) will see empty-string formula results silently converted to blank/null. This is a correctness regression for formulas that intentionally return `""`.

ben-schreiber:
Fixes #356.

## Problem

When the reader encounters a formula cell whose cached result is an empty string — written by Excel/LibreOffice as a self-closing `<v/>` with `t="str"` — the value is dropped and the cell decays to `Empty`, losing its string type. On write the `t="str"` attribute disappears, so `=IF(cond,"","x")` evaluating to `""` becomes an untyped/blank cell. Non-empty string results are unaffected.

Root cause: in `Cell::set_attributes`, the `Event::Start` arm dispatches `<v>` on the declared `t` type, but the `Event::Empty` arm (which is how a self-closing `<v/>` arrives) only handled `<f/>` and ignored `<v/>` entirely.

## Fix

Handle `<v/>` in the `Event::Empty` arm, dispatching on the declared type the same way the `Event::Start` arm does, so an empty string result keeps `t="str"`.

## Repro (before this PR)

`input.xlsx` cell: `<c r="A1" t="str"><f>IF(TRUE,"","x")</f><v/></c>`

```rust
let book = umya_spreadsheet::reader::xlsx::read(Path::new("input.xlsx")).unwrap();
let cell = book.get_sheet(&0).unwrap().get_cell((1, 1)).unwrap();
assert_eq!(cell.get_data_type(), "str"); // was "", now "str"
```

After the fix the written cell keeps its type: `<c r="A1" t="str"><f>IF(TRUE,"","x")</f><v></v></c>`.

## Tests

Added `empty_string_formula_result_keeps_str_type` covering the read path. Full `cargo test --lib` passes (114 tests).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "umya-spreadsheet"
-version = "3.1.0"
+version = "3.1.1"
 authors = ["MathNya <umya.net.info@gmail.com>"]
 repository = "https://github.com/MathNya/umya-spreadsheet"
 keywords = ["excel", "spreadsheet", "xlsx", "reader", "writer"]
--- a/src/structs/cell.rs
+++ b/src/structs/cell.rs
@@ -655,13 +655,32 @@ impl Cell {
                     }
                     _ => (),
                 },
-                Ok(Event::Empty(ref e)) => {
-                    if e.name().into_inner() == b"f" {
+                Ok(Event::Empty(ref e)) => match e.name().into_inner() {
+                    b"f" => {
                         let mut obj = CellFormula::default();
                         obj.set_attributes(reader, e, true, &cell_reference, formula_shared_list);
                         self.cell_value.set_formula_obj(obj);
                     }
-                }
+                    // A self-closing `<v/>` (written by Excel/LibreOffice for an
+                    // empty cached value) still carries the cell's declared type, so
+                    // an empty-string formula result must keep `t="str"` on write.
+                    b"v" => match type_value.as_str() {
+                        "str" => {
+                            self.set_value_string_crate(String::new());
+                        }
+                        "b" => {
+                            self.set_value_bool_crate(false);
+                        }
+                        "e" => {
+                            self.set_error(String::new());
+                        }
+                        "" | "n" => {
+                            self.set_value_crate(String::new());
+                        }
+                        _ => {}
+                    },
+                    _ => (),
+                },
                 Ok(Event::End(ref e)) => if e.name().into_inner() == b"c" { return },
                 Ok(Event::Eof) => panic!("Error: Could not find {} end element", "c"),
                 Err(e) => panic!("Error at position {}: {:?}", reader.buffer_position(), e),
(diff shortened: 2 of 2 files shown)


MathNya:
@ben-schreiber 
Thank you for the PR.
There’s an error in Clippy, but I’ll take care of that.

PR #358, opened 2026-08-28
fix: keep worksheet drawing r:id in sync with its rels

LeandroMarceddu:
write_print_settings bumped the relationship id whenever PageSetup had any page-setup attribute (paperSize/orientation), but the rels writer only emits a printerSettings relationship when PageSetup::object_data is present. For a sheet with a page setup but no printer-settings binary this wrote <drawing r:id="rId2"> while the rels only defined the drawing at rId1, a dangling reference that made Excel strip the drawing. Read the mutable counter back from PageSetup::write_to so the id only advances when a printerSettings relationship is actually written.


MathNya:
@LeandroMarceddu 
Thank you for the PR.

PR #359, opened 2026-08-29
fix: preserve row thickTop

developer0hye:
## Summary

- store the OOXML row `thickTop` flag alongside `thickBot`
- expose v3 `thick_top` / `set_thick_top` accessors and the deprecated compatibility getter
- preserve the flag through XLSX read/write round trips

## Verification

- `cargo test --test integration_test row_thick_top_roundtrips -- --exact`
- `cargo test --lib` (114 passed)
- `cargo +nightly fmt --check`


MathNya:
@developer0hye 
Thank you for the PR.

Issue #184, opened 2024-03-25
Getting the background color when themed.


taylorh140:
I have been finding that argb colors are sometimes empty when getting the background color:

i have this method to get them so far (extracting to json)

```rust
            if let Some(background_color) = cell.get_style().get_background_color() {
                if background_color.get_indexed() != &0  {
                    tmp_cell_data["fill_color"] = json!(INDEXED_COLORS.get(*background_color.get_indexed() as usize));
                } else if background_color.get_theme_index() != &0 {
                    // It would be more useful as argb
                    tmp_cell_data["theme_index"] = json!(*background_color.get_theme_index());
                } else {
                    tmp_cell_data["fill_color"] = json!(background_color.get_argb().to_string());
                }
            }
```

I have to make a copy of the INDEXED_COLORS to extract them at the moment. And im not sure how to decode a themed color.

MathNya:
@taylorh140 
Perhaps the following process can be used to obtain the color code.
```rust
    let book = umya_spreadsheet::reader::xlsx::read(path).unwrap();
    let color = book.get_theme().get_theme_elements()
    .get_color_scheme()
    .get_color_map()
    .get(theme_index)
```
We will fix it so that arbg will always have a numerical value in the next version.

taylorh140:
@MathNya This is super helpful. I'll see if I can get it to work on my end but that looks sort of like the direction I was heading. 

And yeah argb always having a numerical value would be super nice and very helpful in my case. 

Just fyi I have been working on a typst (its similar to a latex)  importing plugin for xlsx sheets. 
so far i can import values and merged cells and i should be able to handle the background colors now.

i think for the first release i would like to get (simple) text alignment and basic borders as well.

but your library is a key technology to make it possible.

Thanks!

taylorh140:
Hey, Is this working? now to get argb?

taylorh140:
@MathNya 

I noticed when i used the solution above i get the darkest version of the themed color:

![image](https://github.com/MathNya/umya-spreadsheet/assets/10996543/8a72680b-bf51-47f4-892b-4d43a8dd562f)

so here i expect :
![image](https://github.com/MathNya/umya-spreadsheet/assets/10996543/8e03c539-e53d-4393-9e04-3e5591928d95)
but end up with:
![image](https://github.com/MathNya/umya-spreadsheet/assets/10996543/5b1fdc25-8253-484d-9d14-0ba616797b9d)

MathNya:
[...]

PR #360, opened 2026-08-29
Fix Excel theme tint color matching

developer0hye:
## Problem

`calc_tint` converts theme colors through a floating-point 255-step HLS space. Excel-compatible exports use the 240-step integer HLS model, so several theme tints resolve one or more RGB levels away from native output. For example, `DAB6BA` at tint `0.7999` resolves to `F8F0F1` instead of `F8EFF0`.

Excel-compatible workbooks also encode common whole-percent tints as nearby signed-short values such as `0.79998168889431442`. Treating those values literally can move the resulting color another level.

## Fix

- convert theme tints through the 240-step integer HLS space
- preserve the integer rounding stages and truncate the tint-adjusted luminance before converting back to RGB
- recover near-whole-percent tint values only within one signed-short step while preserving arbitrary values such as `0.7999`
- update the existing issue 184 fixture expectations to colors measured from a current native Excel PDF export
- add regression coverage for native tint samples, encoded percentages, the normalization threshold, and luminance truncation

## Testing

- `cargo +nightly fmt --all --check`
- `cargo test` (116 unit, 113 integration + 2 ignored, 4 streaming + 1 ignored, 79 doc + 20 ignored)

Related: #184
Related: developer0hye/office2pdf#1394


developer0hye:
CI note: the formatting job passed. The Rust 1.88 job stops during fresh dependency resolution, before this PR's code is compiled:

```text
Adding aes v0.9.3 (requires Rust 1.89)
error: rustc 1.88.0 is not supported by aes@0.9.3
```

This PR does not change `Cargo.toml` or dependency constraints. The remaining build matrix entries were cancelled by fail-fast after the 1.88 job failed. At head `727fe9d4`, `cargo fmt --check`, `cargo clippy -- -D warnings`, and the full local test suite pass (116 unit, 113 integration with 2 ignored, 4 streaming with 1 ignored, and 79 doctests with 20 ignored).

MathNya:
@developer0hye 
To be honest, I had given up on this issue.
Thank you for the PR.
I’ll go ahead and merge it.

PR #361, opened 2026-08-30
Fix literal-only zero number format sections

developer0hye:
## Summary

- render selected number-format sections that contain only literals
- support quoted text, backslash escapes, spacing markers, fill characters, and empty zero sections
- preserve the existing date and numeric formatting paths whenever a value placeholder is present

## Problem

A three-section format such as `#,##0_);[Red]\(#,##0\);\-\ \ ` selects the third section for zero, but the formatter currently falls through to numeric formatting and returns `0` instead of the declared `-` literal.

## Testing

- `cargo test`

The width-independent string formatter represents `*x` with one fill character; it does not expand the character to a cell width.

### DIFF
--- a/src/helper/number_format.rs
+++ b/src/helper/number_format.rs
@@ -88,19 +88,24 @@ pub fn to_formatted_string<S: AsRef<str>, P: AsRef<str>>(value: S, format: P) ->
         _ => {}
     }
 
-    // Convert any other escaped characters to quoted strings, e.g. (\T to "T")
-
-    let mut format = get_escape_regex().replace_all(&format, r#""$0""#);
-
-    // Get the sections, there can be up to four sections, separated with a
-    // semi-colon (but only if not a quoted literal)
-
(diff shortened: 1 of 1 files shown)


developer0hye:
CI note: the failing run is blocked by current baseline/toolchain drift rather than this change.

- `build (1.88.0)` resolves `aes 0.9.3`, which now requires Rust 1.89, because the library does not commit a lockfile.
- `formatting` reports current-nightly wrapping changes across existing files (`date.rs`, `reader/driver.rs`, `writer/streaming_writer.rs`, and others). The reported `number_format.rs` hunk is also an unchanged pre-existing comment.
- The exact PR head passed `cargo test`, `cargo clippy -- -D warnings`, and `cargo +nightly fmt --all -- --check` locally before push; the run logs show dependency/nightly changes after that validation.

No production failure from the literal-only section implementation appears in the CI log.

MathNya:
@developer0hye 
Thank you for the PR.

PR #362, opened 2026-09-03
Preserve whitespace in literal-only number format sections

developer0hye:
## Summary

- preserve leading and trailing whitespace decoded from literal-only number-format sections
- keep escaped spaces in zero-display formats so right-aligned output retains Excel spacing
- extend the #361 regression to assert the complete formatted string

## Problem

#361 correctly selects and decodes a placeholder-free zero section, but the early return trims its result. For `#,##0_);[Red]\(#,##0\);\-\ \ ` that changes the requested `-  ` display to `-`, shifting a right-aligned dash by the width of two spaces.

## Testing

- `cargo test --lib` (119 passed)
- `git diff --check`

Follow-up to #361.

### DIFF
--- a/src/helper/number_format.rs
+++ b/src/helper/number_format.rs
@@ -97,7 +97,7 @@ pub fn to_formatted_string<S: AsRef<str>, P: AsRef<str>>(value: S, format: P) ->
     let raw_sections: Vec<&str> = split(get_section_regex(), &format).collect();
     let (_, raw_split_format, _) = split_format(raw_sections, parsed_val);
     if let Some(literal) = literal_only_section(&raw_split_format) {
-        return literal.trim().to_string();
+        return literal;
     }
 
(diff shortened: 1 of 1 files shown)


developer0hye:
CI is currently red for failures already present on the base branch, rather than for this two-line change:

- `build (1.88.0)` resolves `aes 0.9.3`, which requires Rust 1.89. The repository does not pin a lockfile for this job.
- `formatting` reports existing rustfmt drift across unrelated files such as `src/helper/date.rs`, `src/reader/driver.rs`, and `src/writer/streaming_writer.rs`.

The same base-wide failures are visible on the recently merged #361 and on the upstream `master` workflow. Locally, the focused regression is red before the change and green after it; `cargo test --lib` passes all 119 tests, and `cargo clippy -- -D warnings` passes.

MathNya:
Thank you for the PR.
I'll go ahead and merge it.

Issue #368, opened 2026-09-20
Handle CDATA in raw XML text content


developer0hye:
Reading an XLSX header or footer containing valid CDATA can panic in `helper::utils::unescape_xml_text` on current master (35bbffa4). For example:

```xml
<oddHeader><![CDATA[&C&BResearch & Development&B]]></oddHeader>
```

`Reader::read_text_into` returns the raw inner XML, and the helper unescapes the entire string. It interprets literal CDATA ampersands as entity references and panics with `UnterminatedEntity`. CDATA without ampersands instead retains its XML delimiters.

Expected: CDATA content is literal text, while entity/character references outside CDATA are decoded. Mixed content and adjacent/empty CDATA sections should work, and escaped text resembling a CDATA delimiter should remain ordinary text.

A synthetic workbook regression reproduces this through the public XLSX reader on both header and footer fields. I have a small shared-helper fix with regression coverage in progress.

Related: developer0hye/office2pdf#1709

PR #369, opened 2026-09-20
Fix CDATA decoding in XML text content

developer0hye:
## Summary

Preserve CDATA content when reading XML text with `Reader::read_text_into`. The shared decoder now copies CDATA sections literally and unescapes only the surrounding text, preventing valid header/footer ampersands from panicking and avoiding literal CDATA delimiters in the parsed value.

Related: #368
Related: developer0hye/office2pdf#1709

## Regression coverage

A synthetic workbook exercises both headers and footers through the public XLSX reader: ordinary entities, literal ampersands and tags, mixed text/CDATA, numeric references, adjacent/empty sections, and escaped text that resembles a CDATA delimiter. The test panics on unchanged master and passes with the patch.

## Validation

- `cargo test --locked --test cdata_text`: passed.
- `cargo test --locked`: full suite passed, including documentation tests.
- `cargo clippy --locked -- -D warnings`: passed.
- Targeted nightly rustfmt and `git diff --check`: passed.
- Read-only documentation freshness audit: PASS; no public API or usage change.

The local generated lockfile is retained for repeatability and not committed, following the repository's library policy. No dependencies or binary fixtures are added. This is an AI-assisted implementation.


MathNya:
@developer0hye 
Thank you for the PR.
Since there were no issues with the changes, I will merge them.

Issue #370, opened 2026-09-20
Prefixed worksheet elements silently lose cells and header/footer content


developer0hye:
## Problem

On master (`35bbffa4`, version 3.1.1), prefixing worksheet elements with an alias for the SpreadsheetML namespace causes cells and header/footer content to disappear during reading. The workbook-relationships namespace fix in #367 does not cover worksheet XML.

The synthetic sources below differ only in `xl/worksheets/sheet1.xml`: one uses the default namespace, and the other uses `s:` for the same namespace URI. Expanded element names, attributes and text are identical; all other ZIP parts are byte-identical.

```xml
<s:worksheet xmlns:s="http://schemas.openxmlformats.org/spreadsheetml/2006/main">
  <s:sheetData>
    <s:row r="1"><s:c r="A1" t="inlineStr"><s:is><s:t>Quarterly report</s:t></s:is></s:c></s:row>
  </s:sheetData>
</s:worksheet>
```

`reader::xlsx::read_reader(..., true)` succeeds, but A1 is absent for the prefixed worksheet. The worksheet reader and its nested readers compare literal qualified names such as `row` and `c`.

## Reproduction

- [Default namespace workbook](https://raw.githubusercontent.com/developer0hye/office2pdf/b390661fb9048a83cbd39cf8dbe1bd2e1a0965f4/tests/visual_audits/issue-1803/control.xlsx)
- [Prefixed worksheet workbook](https://raw.githubusercontent.com/developer0hye/office2pdf/b390661fb9048a83cbd39cf8dbe1bd2e1a0965f4/tests/visual_audits/issue-1803/prefixed.xlsx)

Microsoft Excel 16.113 exports both workbooks identically (zero decoded-pixel difference at 150 DPI). In office2pdf the prefixed source becomes a blank page despite reported conversion success:

![Excel left; converter right](https://raw.githubusercontent.com/developer0hye/office2pdf/b390661fb9048a83cbd39cf8dbe1bd2e1a0965f4/assets/bugfixes/issue-1803/compare.jpg)

An in-memory regression generated through the existing writer confirms that the default-namespace control passes while `s:` and `spreadsheet:` worksheet variants fail on this upstream base. It checks text and numeric cells, bold formatting and header/footer content.

Expected: equivalent worksheet namespace prefixes preserve content in normal, lazy and streaming reads.

Downstream tracking: developer0hye/office2pdf#1803.

PR #371, opened 2026-09-20
Resolve worksheet namespace prefixes in streaming readers

developer0hye:
## Summary

Resolve SpreadsheetML namespace prefixes before dispatching worksheet elements to the existing sub-readers. A worksheet using `s:row`, `s:c` and prefixed header/footer elements currently reads successfully but loses its content. The equivalent default-namespace worksheet works.

Related: #370, developer0hye/office2pdf#1803.

A streaming `NsReader` adapter presents the existing sub-readers with their expected unqualified SpreadsheetML names. It resolves each element in scope, retains qualified foreign names, and gives foreign-default-namespace elements an internal prefix so their local names cannot be confused with worksheet elements. The original stored XML is unchanged. Normal, lazy/lite and callback-streaming reads use the adapter. No dependencies or public APIs change.

## Tests

- Red-first synthetic workbook tests: default namespace passes on `35bbffa4`; both `s:` and `spreadsheet:` variants lose A1.
- Text/numeric cells, bold formatting, header/footer text, deferred in-memory reads, lazy file reads, cell streaming, and write/reopen round trips.
- Nested aliases and prefix rebinding, foreign default namespaces, entity/CDATA preservation, and bounded first-event buffering on a 20,000-row XML input.
- `cargo test --locked`: 347 passed; 23 existing ignored tests.
- `cargo clippy --locked --test worksheet_namespaces -- -D warnings`: passed, including the library.
- `cargo +nightly-2026-09-11 fmt --all -- --check` and `git diff --check`: passed.

## Performance tradeoff

The adapter preserves streaming memory behavior but adds a namespace-resolution/serialization pass before the existing parser. An optimized public-API probe with 20,000 rows / 40,000 cells measured these median reader times across six fresh-process runs per variant and mode, interleaved in alternating order:

| Read mode | Unchanged upstream | Patch |
| --- | --- | --- |
| Normal | 80.237 ms | 81.509 ms |
| Cell streaming | 53.636 ms | 71.939 ms |

That is approximately 1.6% and 34.1% slower respectively on this fixture. Measurements were on an Apple M2 with 16 GB RAM under variable host load; every repetition, load observation and process resource report was retained. These are host observations, not a performance guarantee. Baseline and patched executables had distinct SHA-256 hashes; the baseline loses all cells on the prefixed benchmark while the patch preserves all 40,000.

Using an adapter keeps namespace handling consistent through nested worksheet readers without changing their shared `Reader` signatures. The additional streaming-read cost is a tradeoff of that bounded approach.


MathNya:
@developer0hye 
Thank you for the PR.
I'll review the changes. It will take a little while.

Issue #372, opened 2026-09-20
Explicit zero page margins are replaced with defaults on write


developer0hye:
`PageMargins::write_to` uses an all-numeric-zero check to select Excel defaults. Setting all six margins explicitly to zero therefore loses those values when writing and reopening the workbook:

```rust
let mut book = umya_spreadsheet::new_file();
book.sheet_mut(0).unwrap().cell_mut("A1").set_value("Borderless report");
book.sheet_mut(0).unwrap().page_margins_mut()
    .set_left(0.0).set_right(0.0).set_top(0.0)
    .set_bottom(0.0).set_header(0.0).set_footer(0.0);
let mut bytes = Vec::new();
umya_spreadsheet::writer::xlsx::write_writer(&book, &mut bytes).unwrap();
let reopened = umya_spreadsheet::reader::xlsx::read_reader(
    std::io::Cursor::new(bytes), true,
).unwrap();
assert_eq!(reopened.sheet(0).unwrap().page_margins().left(), 0.0);
```

On current master `35bbffa4`, this assertion gets `0.7`; the six values reopen as `[0.7, 0.7, 0.75, 0.75, 0.3, 0.3]` instead of zeros. A workbook whose margins were never set should keep the default behavior.

`DoubleValue` already records presence internally, but `PageMargins` exposes only numeric accessors, so consumers also cannot distinguish absent margin metadata from explicit zero. Presence accessors and using that same state for writer defaults would avoid both ambiguities.

Related downstream report: developer0hye/office2pdf#1812. Synthetic in-memory reproduction; no external files required.

PR #373, opened 2026-09-20
Preserve explicit zero page margins and expose presence

developer0hye:
Explicitly setting all six print margins to zero currently writes Excel defaults instead. This change distinguishes unset margins from supplied values, so zero survives a write/read cycle and defaults apply only to unset fields.

- Add `PageMargins::has_left`, `has_right`, `has_top`, `has_bottom`, `has_header`, and `has_footer` using the existing internal presence state. Existing numeric accessors retain their behavior.
- Select writer defaults independently for each unset margin; partially configured objects now default their unset fields rather than writing numeric zero for them.
- Cover explicit all-zero margins, completely unset margins, mixed zero/nonzero/unset margins, and a worksheet without a `pageMargins` element. Tests use synthetic workbooks generated in memory.
- Document the presence API and writer behavior.

The base-compatible zero-margin regression fails on master `35bbffa4` (all six values reopen as defaults) and passes with this change. The default-only control passes on both.

Validation on macOS:

- `cargo test --locked`: 344 passed, 23 existing ignored tests.
- `cargo clippy --locked -- -D warnings`: passed.
- `cargo clippy --locked --test page_margin_presence -- -D warnings`: passed.
- `cargo +nightly-2026-09-11 fmt --all --check`: passed.
- `git diff --check`: passed.

The local lockfile is untracked, consistent with this repository. No dependency changes or external fixtures.

Related: #372. Downstream: developer0hye/office2pdf#1812.


MathNya:
@developer0hye 
Thank you for the PR.
Since there were no issues with the changes, I will merge them.

